PR #1, opened 2026-04-29
refactor: changing dir of conf and logs

saptarshi-datazip:
Tested by building the new image with both helm and docker compose(locally using jdbc).

### DIFF
--- a/dist/src/main/amoro-bin/bin/load-config.sh
+++ b/dist/src/main/amoro-bin/bin/load-config.sh
@@ -29,11 +29,19 @@ else
 fi
 
 if [ -z "$AMORO_LOG_DIR" ]; then
-    export AMORO_LOG_DIR="${AMORO_HOME}/logs"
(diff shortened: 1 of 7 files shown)


dist/src/main/amoro-bin/bin/load-config.sh
```diff
@@ -29,11 +29,19 @@ else
...
+    fi
 fi
 
 if [ -z "$AMORO_CONF_DIR" ]; then
-    export AMORO_CONF_DIR="${AMORO_HOME}/conf"
+    if [ -n "$FUSION_CONF_DIR" ]; then
+        export AMORO_CONF_DIR="$FUSION_CONF_DIR"
+    else
+        export AMORO_CONF_DIR="${AMORO_HOME}/conf"
+    fi
```
saptarshi-datazip, line 44: can be removed if this is picked up from entrypoint.sh. or can be kept as a fallback.

PR #4, opened 2026-05-22
feat(ams): add startup optimizer bootstrap with auto-recreate optimizer

arsalan-datazip:
Move Spark optimizer setup from Helm lifecycle hooks and Docker Compose init scripts into Fusion itself. On startup, Fusion creates and manages the optimizer resource group and Spark job from optimizer.* config. 

## How was this patch tested?


badalprasadsingh:
Can you please think of these edge cases, and tell me if the solution will work out or not:

- `Spark driver pod evicted/OOM-killed` -> a replacement will be scheduled, the new pod will have the same config, and when it `authenticates`, the old token will still be in DB. Thus, two optimizers should pop up on the UI.

amoro-ams/src/main/java/org/apache/amoro/server/bootstrap/OptimizerBootstrap.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+  }
+
+  private void createOptimizerResource(
+      AbstractOptimizerContainer container, ResourceGroup group, int parallelism) {
+    Resource resource =
+        new Resource.Builder(group.getContainer(), group.getName(), ResourceType.OPTIMIZER)
+            .setProperties(group.getProperties())
+            .setThreadCount(parallelism)
+            .build();
```
badalprasadsingh, line 219: Missed `.setMemorySize(...)`.

amoro-ams/src/main/java/org/apache/amoro/server/bootstrap/OptimizerBootstrap.java
```diff
@@ -0,0 +1,190 @@
...
+          "Cannot release optimizer on non-optimizer resource container %s.",
+          resource.getContainerName());
+      ((AbstractOptimizerContainer) resourceContainer).releaseResource(resource);
+      released = true;
+    } catch (Exception e) {
+      LOG.warn("Failed to release optimizer resource {}", resourceId, e);
+    }
+    if (released) {
+      cleanupResource(resource.getGroupName(), resourceId);
+    }
```
badalprasadsingh, line 260: Here, the `finally` block always runs. Are we sure it is required to behave like this? What if the kill fails?

amoro-ams/src/main/java/org/apache/amoro/server/bootstrap/OptimizerBootstrap.java
```diff
@@ -0,0 +1,190 @@
...
+
+  private void createOptimizerResource(
+      AbstractOptimizerContainer container, ResourceGroup group, int parallelism) {
+    Resource resource =
+        new Resource.Builder(group.getContainer(), group.getName(), ResourceType.OPTIMIZER)
+            .setProperties(group.getProperties())
+            .setThreadCount(parallelism)
+            .build();
+    container.requestResource(resource);
+    optimizerManager.createResource(resource);
```
badalprasadsingh, line 221: Any reason to request the resource before creating the resource?

amoro-ams/src/main/java/org/apache/amoro/server/bootstrap/OptimizerBootstrap.java
```diff
@@ -0,0 +1,190 @@
...
+    } catch (Exception e) {
+      LOG.warn("Failed to delete resource {} from database", resourceId, e);
+    }
+    try {
+      optimizerManager.deleteOptimizer(groupName, resourceId);
+    } catch (Exception e) {
+      LOG.warn("Failed to delete optimizer {} from database during cleanup", resourceId, e);
+    }
+  }
+}
```
badalprasadsingh, line 282: Isn't this redundant?

amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
```diff
@@ -257,7 +260,7 @@ public void startOptimizingService() throws Exception {
     tableService.initialize();
     LOG.info("AMS table service have been initialized");
     tableManager.setTableService(tableService);
-
+    new OptimizerBootstrap(serviceConfig, optimizerManager, optimizingService).run();
```
badalprasadsingh: Please add this into a try-catch block.

badalprasadsingh (changes requested):
First Review.

Please add the modification header where-ever required.

amoro-ams/src/main/java/org/apache/amoro/server/bootstrap/OptimizerBootstrap.java
arsalan-datazip: fixed

amoro-ams/src/main/java/org/apache/amoro/server/bootstrap/OptimizerBootstrap.java
arsalan-datazip: requestResource first: doScaleOut enriches resource.getProperties() with metadata
needed by releaseResource. 
createResource then persists the complete record
before the optimizer's authenticate() API call arrives.

This is also followed while creating the resource thorugh API.

amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
arsalan-datazip: done

amoro-ams/src/main/java/org/apache/amoro/server/bootstrap/OptimizerBootstrap.java
arsalan-datazip: done changes

badalprasadsingh:
Closing this **PR** as the _direction_ of the feature has moved to _auto-creation_ of optimizer as well

EDIT: Let's not create a new PR

badalprasadsingh:
Closing this PR as the feature direction has moved a lot.

PR #6, opened 2026-06-02
feat: test catalog connection

arsalan-datazip:
This PR adds a catalog connection test API that will validate catalog credentials/configuration before creating the catalog.

## How was this patch tested?


amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogTableListing.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Comparator;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/** Shared table-listing logic used by the dashboard API and catalog connection tests. */
+public final class CatalogTableListing {
+
+  private CatalogTableListing() {}
```
badalprasadsingh: Any specific reason for making the constructor private?

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/model/CatalogConnectionTestResult.java
```diff
@@ -0,0 +1,47 @@
...
+    this.connected = connected;
+  }
+
+  public String getMessage() {
+    return message;
+  }
+
+  public void setMessage(String message) {
+    this.message = message;
+  }
```
badalprasadsingh: Are these setters and getters being used somewhere? If not, please remove them.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/CatalogController.java
```diff
@@ -601,16 +602,34 @@ private void unMaskSensitiveData(CatalogRegisterInfo newInfo, CatalogMeta oldCat
...
+    CatalogMeta catalogMeta = parseCatalogInfo(ctx);
+    if (catalogService.catalogExist(catalogMeta.getCatalogName())) {
       throw new RuntimeException("Duplicate catalog name!");
     }
-    CatalogMeta catalogMeta = constructCatalogMeta(info, null);
     catalogService.createCatalog(catalogMeta);
     ctx.json(OkResponse.of(""));
   }
 
+  /** Test catalog connection without persisting catalog metadata. */
```
badalprasadsingh: Just keep "Test catalog connection"

amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogManager.java
```diff
@@ -100,4 +101,12 @@ public interface CatalogManager {
    * @return managed table.
    */
   AmoroTable<?> loadTable(TableIdentifier identifier);
+
+  /**
+   * Test catalog connectivity without persisting catalog metadata.
```
badalprasadsingh: Remove " without persisting catalog metadata."

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
```diff
@@ -240,6 +241,35 @@ public AmoroTable<?> loadTable(TableIdentifier identifier) {
...
+      }
+      return new CatalogConnectionTestResult(true, "Test connection successful");
+    } catch (Exception e) {
+      LOG.warn(
+          "Catalog connection test failed for {}: {}",
+          catalogMeta.getCatalogName(),
+          e.getMessage(),
+          e);
+      return new CatalogConnectionTestResult(
+          false, e.getMessage() != null ? e.getMessage() : e.toString());
```
badalprasadsingh: Please add a proper message of Test Connection Unsuccessful

badalprasadsingh:
This test connection feature won't work when editing a catalog. Please check it once.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
```diff
@@ -240,6 +241,35 @@ public AmoroTable<?> loadTable(TableIdentifier identifier) {
...
+  @Override
+  public CatalogConnectionTestResult testCatalogConnection(CatalogMeta catalogMeta) {
+    ServerCatalog catalog = null;
+    try {
+      fillCatalogProperties(catalogMeta);
+      catalog = CatalogBuilder.buildServerCatalog(catalogMeta, serverConfiguration);
+      List<String> databases = catalog.listDatabases();
+      if (!databases.isEmpty()) {
+        String testDatabase =
+            databases.stream().sorted().findFirst().orElseThrow(IllegalStateException::new);
```
badalprasadsingh: Is the `IllegalStateException` case ever reachable? as we already have the `isEmpty()` check beforehand

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/model/CatalogConnectionTestResult.java
```diff
@@ -0,0 +1,47 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.amoro.server.dashboard.model;
+
+public class CatalogConnectionTestResult {
```
badalprasadsingh: No point of making it a public DTO. Make the class final.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/model/CatalogConnectionTestResult.java
```diff
@@ -0,0 +1,47 @@
...
+  private boolean connected;
+  private String message;
```
badalprasadsingh: Make the fields final as well. This is a response, so let it be immutable.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/model/CatalogConnectionTestResult.java
```diff
@@ -0,0 +1,47 @@
...
+
+  public CatalogConnectionTestResult() {}
```
badalprasadsingh: Once you make the class final, remove this constructor as well

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/CatalogController.java
```diff
@@ -601,16 +602,34 @@ private void unMaskSensitiveData(CatalogRegisterInfo newInfo, CatalogMeta oldCat
...
   /** Register catalog to ams. */
   public void createCatalog(Context ctx) {
-    CatalogRegisterInfo info = ctx.bodyAsClass(CatalogRegisterInfo.class);
-    validateCatalogRegisterInfo(info);
-    if (catalogService.catalogExist(info.getName())) {
+    CatalogMeta catalogMeta = parseCatalogInfo(ctx);
+    if (catalogService.catalogExist(catalogMeta.getCatalogName())) {
       throw new RuntimeException("Duplicate catalog name!");
     }
-    CatalogMeta catalogMeta = constructCatalogMeta(info, null);
```
badalprasadsingh, line 609: Previously, we use to check `catalogService.catalogExist(info.getName())`, then only call the `constructCatalogMeta` function.

Now with your refactoring, this will be called everytime, even for duplicates. Please check the consequences of this.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
```diff
@@ -240,6 +241,35 @@ public AmoroTable<?> loadTable(TableIdentifier identifier) {
...
+      LOG.warn(
+          "Catalog connection test failed for {}: {}",
+          catalogMeta.getCatalogName(),
+          e.getMessage(),
+          e);
+      return new CatalogConnectionTestResult(
+          false, e.getMessage() != null ? e.getMessage() : e.toString());
+    } finally {
+      if (catalog != null) {
+        catalog.dispose();
```
badalprasadsingh: The dispose function implementation is empty. What's the point of calling it?

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
```diff
@@ -240,6 +241,35 @@ public AmoroTable<?> loadTable(TableIdentifier identifier) {
     return serverCatalog.loadTable(identifier.getDatabase(), identifier.getTableName());
   }
 
+  //  TODO: create a test_olake namespace and test_olake table to test the connection
+  @Override
+  public CatalogConnectionTestResult testCatalogConnection(CatalogMeta catalogMeta) {
+    ServerCatalog catalog = null;
+    try {
+      fillCatalogProperties(catalogMeta);
+      catalog = CatalogBuilder.buildServerCatalog(catalogMeta, serverConfiguration);
```
badalprasadsingh: `catalogMeta` is updated in the `fillCatalogProperties` function.

Please make a deep copy of it this way, and use that.
```java
CatalogMeta copy = catalogMeta.deepCopy(); 
```

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/CatalogController.java
```diff
@@ -601,16 +602,34 @@ private void unMaskSensitiveData(CatalogRegisterInfo newInfo, CatalogMeta oldCat
...
     catalogService.createCatalog(catalogMeta);
     ctx.json(OkResponse.of(""));
   }
 
+  /** Test catalog connection without persisting catalog metadata. */
+  public void testConnection(Context ctx) {
```
badalprasadsingh: Check what is the timeout of this test Connection API. If there is no end timeout please keep it `10 mins`.

Reason: even the check destination api in olake-ui has a maximum timeout set as 10 mins

badalprasadsingh (changes requested):
Reviewed. Please add the `modification-header` wherever required.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogTableListing.java
arsalan-datazip: Because we are not creating any object using this class.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/model/CatalogConnectionTestResult.java
arsalan-datazip: Removed

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/CatalogController.java
arsalan-datazip: I saw a timeout of 45000ms once in the UI, do we need to change that to 10 mins

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/CatalogController.java
```diff
@@ -602,15 +605,29 @@ private void unMaskSensitiveData(CatalogRegisterInfo newInfo, CatalogMeta oldCat
   /** Register catalog to ams. */
   public void createCatalog(Context ctx) {
     CatalogRegisterInfo info = ctx.bodyAsClass(CatalogRegisterInfo.class);
-    validateCatalogRegisterInfo(info);
-    if (catalogService.catalogExist(info.getName())) {
-      throw new RuntimeException("Duplicate catalog name!");
-    }
```
badalprasadsingh, line 608: Please don't remove it now. I understand that it is redundant check both in TestConnection and in CreateCatalog.

But, here is the edge case, right now we also use Amoro UI for debugging. We don't do test catalog connection there. So until we refactor the whole codebase, let it be there.

Let it be as it is.

```java
public void createCatalog(Context ctx) {
    CatalogRegisterInfo info = ctx.bodyAsClass(CatalogRegisterInfo.class);
    validateCatalogRegisterInfo(info);
    if (catalogService.catalogExist(info.getName())) {
      throw new RuntimeException("Duplicate catalog name!");
    }
```

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
```diff
@@ -240,6 +242,33 @@ public AmoroTable<?> loadTable(TableIdentifier identifier) {
...
+          "Catalog connection test failed for {}: {}",
+          catalogMeta.getCatalogName(),
+          e.getMessage(),
+          e);
+      String errorMsg =
+          "Test Connection unsuccessful: "
+              + (e.getMessage() != null ? e.getMessage() : e.toString());
+      return new CatalogConnectionTestResult(false, errorMsg);
+    }
+  }
```
badalprasadsingh: Even though I last checked out the "dispose" function is empty, I see in `createCatalog` function they do :

```java
disposeCatalog(catalogMeta.getCatalogName());
```

Lets do this in TestConnection as well.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogTableListing.java
```diff
@@ -0,0 +1,92 @@
...
+    List<TableMeta> tables =
+        new ArrayList<>(
+            serverCatalog.listTables(database).stream()
+                .map(
+                    idWithFormat ->
+                        new TableMeta(
+                            idWithFormat.getIdentifier().getTableName(),
+                            formatToType.apply(idWithFormat.getTableFormat())))
+                .sorted(Comparator.comparing(TableMeta::getType).thenComparing(TableMeta::getName))
+                .collect(Collectors.toList()));
```
badalprasadsingh:
```java
            .sorted(
                (table1, table2) -> {
                  if (Objects.equals(table1.getType(), table2.getType())) {
                    return table1.getName().compareTo(table2.getName());
                  } else {
                    return table1.getType().compareTo(table2.getType());
                  }
                })
            .collect(Collectors.toList());
```

amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogTableListing.java
```diff
@@ -0,0 +1,92 @@
...
+
+    String catalogType = serverCatalog.getMetadata().getCatalogType();
+    if (Objects.equals(catalogType, CATALOG_TYPE_HIVE)) {
```
badalprasadsingh:
```java
catalogType.equals(CATALOG_TYPE_HIVE)
```

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/CatalogController.java
```diff
@@ -602,15 +605,29 @@ private void unMaskSensitiveData(CatalogRegisterInfo newInfo, CatalogMeta oldCat
...
     ctx.json(OkResponse.of(""));
   }
 
+  // test catalog connnection
+  public void testConnection(Context ctx) {
+    try {
+      CatalogRegisterInfo info = ctx.bodyAsClass(CatalogRegisterInfo.class);
+      if (catalogService.catalogExist(info.getName())) {
+        throw new RuntimeException("Duplicate catalog name!");
+      }
```
badalprasadsingh: How will Test Catalog Connection work in Edit Catalog scenario?

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/CatalogController.java
badalprasadsingh: Thus, adding the catalog Exist check in Create Catalog, as Test Catalog connection should also be used for Edit Catalog, and it is missing right now.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/CatalogController.java
```diff
@@ -602,15 +605,29 @@ private void unMaskSensitiveData(CatalogRegisterInfo newInfo, CatalogMeta oldCat
...
-    validateCatalogRegisterInfo(info);
-    if (catalogService.catalogExist(info.getName())) {
-      throw new RuntimeException("Duplicate catalog name!");
-    }
     CatalogMeta catalogMeta = constructCatalogMeta(info, null);
     catalogService.createCatalog(catalogMeta);
     ctx.json(OkResponse.of(""));
   }
 
+  // test catalog connnection
```
badalprasadsingh: "connection"

amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogManager.java
```diff
@@ -100,4 +103,12 @@ public interface CatalogManager {
    * @return managed table.
    */
   AmoroTable<?> loadTable(TableIdentifier identifier);
+
+  /**
+   * Test catalog connection.
+   *
+   * @param catalogMeta the catalog meta information used to build a temporary catalog
+   * @return connection test result with database list or failure message
```
badalprasadsingh: we never return the database list

badalprasadsingh (changes requested):
Second Review. Please add the test connection logic for Edit Catalog.

Note:
- As you said timeout is 45s right now. We might have to increase it later on.

badalprasadsingh:
Please test with wrong object store credentials with the catalog having atleast one database. This will hit listTables.

Need to confirm if this Test Connection logic fails for such a scenario or not.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogConnectionTester.java
```diff
@@ -0,0 +1,173 @@
...
+  static final String TEST_NAMESPACE = "test_olake";
+  static final String TEST_TABLE = "test_olake";
+
+  private static final Schema TEST_SCHEMA =
+      new Schema(
+          Types.NestedField.required(1, "_olake_id", Types.StringType.get()),
+          Types.NestedField.optional(2, "_olake_timestamp", Types.TimestampType.withZone()),
+          Types.NestedField.required(3, "_op_type", Types.StringType.get()),
+          Types.NestedField.optional(4, "_cdc_timestamp", Types.TimestampType.withZone()),
+          Types.NestedField.required(5, "data", Types.StringType.get()));
```
badalprasadsingh: Check with the test_olake table created during ingestion. All are 4 of them are marked as required false.

Image

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/CatalogController.java
```diff
@@ -611,6 +614,21 @@ public void createCatalog(Context ctx) {
     ctx.json(OkResponse.of(""));
   }
 
+  // test catalog connection
+  public void testConnection(Context ctx) {
+    try {
+      CatalogRegisterInfo info = ctx.bodyAsClass(CatalogRegisterInfo.class);
+      validateCatalogRegisterInfo(info);
+      CatalogMeta catalogMeta = constructCatalogMeta(info, null);
+      ctx.json(OkResponse.of(catalogService.testCatalogConnection(catalogMeta)));
```
badalprasadsingh: While update catalog, fusion masks the credentials (access keys and secret keys). If you look at the UpdateCatalog Implementation of fusion, we call the `unMaskSensitiveData`.

Please test it. IMO, we need to do it here.

badalprasadsingh (changes requested):
Few more comments.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
```diff
@@ -240,6 +241,29 @@ public AmoroTable<?> loadTable(TableIdentifier identifier) {
...
+          "Catalog connection test failed for {}: {}",
+          catalogMeta.getCatalogName(),
+          e.getMessage(),
+          e);
+      String errorMsg =
+          "Test Connection unsuccessful: "
+              + (e.getMessage() != null ? e.getMessage() : e.toString());
+      return new CatalogConnectionTestResult(false, errorMsg);
+    } finally {
+      disposeCatalog(catalogMeta.getCatalogName());
```
badalprasadsingh: If we dispose the catalog using the name, then in update catalog test connection, we lose out on it.

In that case, we might not have to dispose the catalog. Check the example function I gave you.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
```diff
@@ -240,6 +241,29 @@ public AmoroTable<?> loadTable(TableIdentifier identifier) {
     return serverCatalog.loadTable(identifier.getDatabase(), identifier.getTableName());
   }
 
+  @Override
+  public CatalogConnectionTestResult testCatalogConnection(CatalogMeta catalogMeta) {
+    try {
+      CatalogMeta copy = catalogMeta.deepCopy();
+      fillCatalogProperties(copy);
+      ServerCatalog catalog = CatalogBuilder.buildServerCatalog(copy, serverConfiguration);
```
badalprasadsingh: Can you please check if we can avoid creating a serverCatalog? 

Because in Fusion, we have two catalogs: External and Internal. For our use case, external catalogs already have the info the deep copy we are making.

badalprasadsingh: A typical example function should look like this. Test it please.

```java
  @Override
  public CatalogConnectionTestResult testCatalogConnection(CatalogMeta catalogMeta) {
    try {
      CatalogMeta copy = catalogMeta.deepCopy();
      fillCatalogProperties(copy);
      CatalogConnectionTester.runTestConnection(copy);
      return new CatalogConnectionTestResult(true, "Test connection successful");
    } catch (Exception e) {
      LOG.warn(
          "Catalog connection test failed for {}: {}",
          catalogMeta.getCatalogName(),
          e.getMessage(),
          e);
      return new CatalogConnectionTestResult(
          false, "Test Connection unsuccessful: " + ExceptionUtils.getRootCauseMessage(e));
    }
  }
```

amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogConnectionTester.java
```diff
@@ -0,0 +1,173 @@
...
+
+  public static void runTestConnection(ServerCatalog serverCatalog) throws Exception {
```
badalprasadsingh: We might not have to pass serverCatalog. We can simply do it with the catalog meta.

```java
public static void runTestConnection(CatalogMeta catalogMeta)
```

amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogConnectionTester.java
```diff
@@ -0,0 +1,173 @@
...
+            catalogName, metastoreType, catalogMeta.getCatalogProperties());
+    Map<String, String> icebergProps =
+        CatalogMetaProperties.CATALOG_TYPE_GLUE.equalsIgnoreCase(metastoreType)
+            ? StaticAwsCredentialsProvider.applyGlueCredentials(baseIcebergProps)
+            : baseIcebergProps;
+    // create iceberg catalog
+    Catalog catalog =
+        org.apache.iceberg.CatalogUtil.buildIcebergCatalog(
+            catalogName, icebergProps, metaStore.getConfiguration());
+    createNamespaceAndTable(catalog, catalogName);
```
badalprasadsingh: We should wrap this up under metaStore.doAs, you can check the use case of metastore.DoAs inside the codebase.

```java
    metaStore.doAs(
        () -> {
          Catalog catalog =
              org.apache.iceberg.CatalogUtil.buildIcebergCatalog(
                  catalogName, icebergProps, metaStore.getConfiguration());
          try {
            createNamespaceAndTable(catalog, catalogName);
          } finally {
            closeQuietly(catalog);
          }
          return null;
        });
```

badalprasadsingh: You add this function closing the catalog. As we are no longer building a server catalog.

```java
  private static void closeQuietly(Catalog catalog) {
    if (catalog instanceof Closeable) {
      try {
        ((Closeable) catalog).close();
      } catch (Exception e) {
        LOG.warn("Failed to close iceberg catalog after connection test: {}", e.getMessage(), e);
      }
    }
  }
```

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/CatalogController.java
badalprasadsingh: Check this:

```java
  public void testConnection(Context ctx) {
    CatalogRegisterInfo info;
    CatalogMeta catalogMeta;
    try {
      info = ctx.bodyAsClass(CatalogRegisterInfo.class);
      validateCatalogRegisterInfo(info);
      CatalogMeta oldCatalogMeta = null;
      if (catalogService.catalogExist(info.getName())) {
        oldCatalogMeta = catalogService.getCatalogMeta(info.getName());
        unMaskSensitiveData(info, oldCatalogMeta);
      }
      catalogMeta = constructCatalogMeta(info, oldCatalogMeta);
    } catch (Exception e) {
      String errorMsg = "Invalid catalog configuration: " + ExceptionUtils.getRootCauseMessage(e);
      ctx.json(OkResponse.of(new CatalogConnectionTestResult(false, errorMsg)));
      return;
    }
    ctx.json(OkResponse.of(catalogService.testCatalogConnection(catalogMeta)));
  }
```

badalprasadsingh (changes requested):
Please check these.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
arsalan-datazip: > Can you please check if we can avoid creating a serverCatalog?
> 
> Because in Fusion, we have two catalogs: External and Internal. For our use case, external catalogs already have the info the deep copy we are making.

Yes we do not need to create serverCatalog for test connection because buildServerCatalog validates for catalog format and types and in our case we are using only external catalog wiith iceberg table format.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
arsalan-datazip: Yes we do not need to dispose catalog because we are not storing in the DB for test connection, createCatalog do it because it put catalog in the DB and then dispose it.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/CatalogController.java
```diff
@@ -611,6 +614,21 @@ public void createCatalog(Context ctx) {
     ctx.json(OkResponse.of(""));
   }
 
+  // test catalog connection
+  public void testConnection(Context ctx) {
+    try {
+      CatalogRegisterInfo info = ctx.bodyAsClass(CatalogRegisterInfo.class);
+      validateCatalogRegisterInfo(info);
+      CatalogMeta catalogMeta = constructCatalogMeta(info, null);
```
badalprasadsingh: Shouldn't we pass the oldCatalogMeta info we collected? It might break in update catalog.

`constructCatalogMeta(info, oldCatalogMeta)`

badalprasadsingh (dismissed):

amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogConnectionTester.java
```diff
@@ -0,0 +1,188 @@
...
+    TableMetaStore metaStore = CatalogUtil.buildMetaStore(catalogMeta);
+    String metastoreType = catalogMeta.getCatalogType();
+    // We cannot use icebergCatalog.java  because it does not support createTable operation.
+    // Building raw iceberg catalog that supports createTable operation.
+    Map<String, String> baseIcebergProps =
+        MixedFormatCatalogUtil.withIcebergCatalogInitializeProperties(
+            catalogName, metastoreType, catalogMeta.getCatalogProperties());
+    Map<String, String> icebergProps =
+        CatalogMetaProperties.CATALOG_TYPE_GLUE.equalsIgnoreCase(metastoreType)
+            ? StaticAwsCredentialsProvider.applyGlueCredentials(baseIcebergProps)
```
hash-data, line 80: this was added before as well

hash-data (dismissed):

hash-data (approved):

PR #8, opened 2026-06-09
chore: refactored test conn

hash-data:
Pr refactor test connection, on create and update catalog test connection runs.


arsalan-datazip (approved):
LGTM

PR #10, opened 2026-07-02
fix(ams): recover iceberg tables stranded in planning to pending state

badalprasadsingh:
## Description

Due to any reason, for e.g., in case the `AMS` shuts down, and the iceberg table is stored in the database in `PLANNING` state of status code `500`, the iceberg table remains in the `PLANNING` state even after restarting the AMS.

## Solution

Whenever AMS starts, check for any table stranded on code `500` and bring it to `PENDING` state `600`.

Fixes: #11 

## Testing

In `local-test`,

1. Close the setup: `make clean-fusion-docker`
2. Just start the `postgres` service, and set the iceberg table to `500` state:
```bash
docker compose --profile dev up -d postgres
```

```sql
docker exec -it fusion-postgres psql -U iceberg -d iceberg -c "
UPDATE table_runtime
SET status_code = 500
WHERE table_id = (
    SELECT table_id
    FROM table_identifier
    WHERE catalog_name = 'glue'
      AND db_name = 'aloo'
      AND table_name = 'auto_table_9999'
);
"
```
3. And then on, `make start-fusion-docker`. Without the fix, the iceberg table would still remain in the `PLANNING` state, but with the fix, the iceberg table will come to `PENDING` state and ultimately go to `IDLE` state.


amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -372,6 +372,21 @@ public void planFailed() {
...
+  /**
+   * Recovers tables stranded in {@link OptimizingStatus#PLANNING} to {@link
+   * OptimizingStatus#PENDING};
+   */
+  public void recoverOrphanedPlanning() {
+    store()
+        .begin()
+        .updateStatusCode(
+            code ->
+                code == OptimizingStatus.PLANNING.getCode()
```
hash-data: already there is check when the functionis getting called, also there is no such function exist right?

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -140,6 +140,14 @@ public OptimizingQueue(
   }
 
   private void initTableRuntime(DefaultTableRuntime tableRuntime) {
+    // Recover tables stranded in PLANNING to PENDING
```
hash-data, line 143: This function is also get called whenever a new resource group i.e., an optimizer get created. I know in our case it will not affect but can we check if that is a problem (later on there can be a feature to create multiple optimizers)?

badalprasadsingh: I understand the case.

Whenever we create a new `OptimizerGroup`, the `OptimizingQueue` is empty so it is safe, as creating or registering an optimizer does not create a queue filled with tables (its empty).

The case you are talking about is only possible if: Whenever a new `OptimizerGroup` is created, it has an `OptimizingQueue` containing tables that is being planned in another live queue.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
badalprasadsingh: Done.

hash-data (approved):

PR #14, opened 2026-07-14
feat(ams): optimize get tables api

badalprasadsingh:
## Description

Optimize the API of OLake-Fusion side, so OLake-UI for Fusion does not have to make several round trips to get the same fields across several iceberg tables.

1. Get Tables API (Dashboard)
2. Get Optimization Status per Iceberg Table (_Issue_: Everytime, someone enables/disables an iceberg table, UI reloads the whole dashboard, calling `API (1)` to fetch all the iceberg tables - leading to waiting for `1min` again.)

## Testing
1. For `200` Iceberg Tables each of `80GB`, containing almost `400` data files and `100` delete files: It now takes `685 ms`, previously it took around `43.51s`. 

**Note:** From Fusion side, it is backward compatible, From OLake-UI side, frontend has to have the `backward-compatibility` check.


badalprasadsingh:
Closing this PR as the PR #32 will override the requirements of this feature.

PR #15, opened 2026-07-15
feat: telemetry

ashisharma-dz:

Close #xxx.

## Brief change log

Integrated the `Telemetry` tracking client : amoro-ams/src/test/java/org/apache/amoro/server/utils/Telemetry.java
- Tracked basic information including platform and location
- Hooked telemetry trigger points into the api handlers to capture when compactions start and complete in: 
- amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java, 
- Added event triggers for catalog addition events (e.g., catalog additions): amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java

## How was this patch tested?

- [x]Ran instances of compaction and catalog addition,  and confirmed the live logs


amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,281 @@
...
+  private Telemetry() {
+    this.httpClient = HttpClient.newBuilder().connectTimeout(Duration.ofSeconds(5)).build();
+    this.objectMapper = new ObjectMapper();
+
+    initAsync();
+  }
+
+  // =========================================================================
+  // INFRASTRUCTURE & METADATA UTILITIES (Placed before trackers)
+  // =========================================================================
```
badalprasadsingh: Remove this.

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -228,6 +232,16 @@ public OptimizingTask pollTask(String authToken, int threadId) {
...
+    OptimizingType optimizingType = queue.getOptimizingType(taskId);
+    TaskRuntime<?> taskRuntime = queue.getTaskRuntime(taskId);
+    long tableSize = 0;
+    if (taskRuntime.getTaskDescriptor() instanceof RewriteStageTask) {
+      MetricsSummary summary = ((RewriteStageTask) taskRuntime.getTaskDescriptor()).getSummary();
+      if (summary != null) {
+        tableSize = summary.getInputFilesStatistics().getTotalSize();
+      }
+    }
+    Telemetry.getInstance().trackCompactionStarted(optimizingType, tableSize);
```
badalprasadsingh: Here, `getOptimizingType` or `getTaskRuntime` both call `findProcess`, which throws `TaskNotFoundException`. If any error or nil is thrown here, `ackTask` or `completeTask` never runs.

Please add this block in a `try`/`catch` block.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,281 @@
...
+              });
+
+    } catch (Exception e) {
+      LOG.debug("Telemetry delivery structure failure: {}", e.getMessage());
+    }
+  }
+
+  // =========================================================================
+  // EXPLICIT TRACKER TRIGGERS
+  // =========================================================================
```
badalprasadsingh: Remove these from all areas.

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -228,6 +232,16 @@ public OptimizingTask pollTask(String authToken, int threadId) {
   public void ackTask(String authToken, int threadId, OptimizingTaskId taskId) {
     LOG.info("Ack task {} by optimizer {} (threadId {})", taskId, authToken, threadId);
     OptimizingQueue queue = getQueueByToken(authToken);
+    OptimizingType optimizingType = queue.getOptimizingType(taskId);
+    TaskRuntime<?> taskRuntime = queue.getTaskRuntime(taskId);
```
badalprasadsingh: The problem is, we are calling `findProcess` 3 times per `ack`/`complete`. Can we call it one time and reuse it?

Tell me if it is not possible.

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -240,6 +254,18 @@ public void completeTask(String authToken, OptimizingTaskResult taskResult) {
...
+    OptimizingTaskId taskId = taskResult.getTaskId();
+    OptimizingType optimizingType = queue.getOptimizingType(taskId);
+    TaskRuntime<?> taskRuntime = queue.getTaskRuntime(taskId);
+    long tableSize = 0;
+    if (taskRuntime.getTaskDescriptor() instanceof RewriteStageTask) {
+      MetricsSummary summary = ((RewriteStageTask) taskRuntime.getTaskDescriptor()).getSummary();
+      if (summary != null) {
+        tableSize = summary.getInputFilesStatistics().getTotalSize();
+      }
+    }
```
badalprasadsingh: Common out this in a function and reuse. Duplicate code logic above as well in `ackTask`

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,281 @@
...
+  private static double bytesToGb(long bytes) {
+    return bytes / BYTES_PER_GB;
+  }
+
+  public void trackCompactionStarted(OptimizingType compactionType, long tableSize) {
+    CompletableFuture.runAsync(
+        () -> {
+          Map<String, Object> props =
+              Map.of("compaction_type", compactionType.name(), "table_size", bytesToGb(tableSize));
+          sendEvent("Compaction started - Fusion", props);
```
badalprasadsingh: For these events, as they would go to Product side for analytics. Lets follow a standard for these messages. `Compaction` -> `Optimization`, and all words starting with Capital Cases. 

Make it throughout all messages.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,281 @@
...
+  private void sendEvent(String eventName, Map<String, Object> props) {
+    if (platform == null) {
+      return; // Engine initialization was skipped or telemetry is intentionally disabled
+    }
+
+    try {
+      Map<String, Object> enrichedProperties = new HashMap<>(props);
+      enrichedProperties.put("os", platform.os());
+      enrichedProperties.put("arch", platform.arch());
+      enrichedProperties.put("olake_version", platform.version());
```
badalprasadsingh: `oLake-fusion_version`

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,281 @@
...
+      Files.writeString(idPath, generatedID, StandardCharsets.UTF_8);
+      return generatedID;
+    } catch (IOException | NoSuchAlgorithmException e) {
+      LOG.debug("Failed to parse or save telemetry user id file, using transient identity.", e);
+      return "transient-" + java.util.UUID.randomUUID().toString().substring(0, 8);
+    }
+  }
+
+  /** Core abstract event transport processing logic matching Go's (t *Telemetry) sendEvent */
+  private void sendEvent(String eventName, Map<String, Object> props) {
```
badalprasadsingh, line 241: IMO this function is running as double `async`. Please recheck it, understand the consequences, if any major, then we have to change it.

badalprasadsingh (changes requested):
First Review.

1. Catalog Imported from OLake or not is not handled here. Confirm me, if it is going to be sent from OLake-UI Metric
2. Change the base of the PR from `master` to `staging`
3. Change the PR title to `feat: telemetry`

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -240,9 +252,29 @@ public void completeTask(String authToken, OptimizingTaskResult taskResult) {
...
+      OptimizingType optimizingType = queue.getOptimizingType(taskId);
+      TaskRuntime<?> taskRuntime = queue.getTaskRuntime(taskId);
+      Telemetry.getInstance()
+          .trackCompactionCompleted(optimizingType, getInputTableSize(taskRuntime), success);
+      OptimizerThread thread =
+          getAuthenticatedOptimizer(authToken).getThread(taskResult.getThreadId());
+      queue.completeTask(thread, taskResult);
+    } catch (Exception e) {
+      LOG.warn("Failed to track compaction completed for task {}", taskId, e);
+    }
```
badalprasadsingh: Earlier: all the errors were sent back to optimizer, we need to do the same here.

Right now, we just continue by logging just a WARN statement.

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -228,7 +232,15 @@ public OptimizingTask pollTask(String authToken, int threadId) {
   public void ackTask(String authToken, int threadId, OptimizingTaskId taskId) {
     LOG.info("Ack task {} by optimizer {} (threadId {})", taskId, authToken, threadId);
     OptimizingQueue queue = getQueueByToken(authToken);
-    queue.ackTask(taskId, getAuthenticatedOptimizer(authToken).getThread(threadId));
+    try {
+      OptimizingType optimizingType = queue.getOptimizingType(taskId);
+      TaskRuntime<?> taskRuntime = queue.getTaskRuntime(taskId);
+      Telemetry.getInstance()
+          .trackCompactionStarted(optimizingType, getInputTableSize(taskRuntime));
+      queue.ackTask(taskId, getAuthenticatedOptimizer(authToken).getThread(threadId));
```
badalprasadsingh: Ordering issue.

We track compaction started before `ackTask` even succeeds. If `getOptimizingType`/`getTaskRuntime` throws an exception (`TaskNotFoundException`) and fail, `queue.ackTask` is skipped.

Better:
```java
queue.ackTask(taskId, getAuthenticatedOptimizer(authToken).getThread(threadId));
try {
  Telemetry.getInstance().trackCompactionStarted(
      queue.getOptimizingType(taskId), getInputTableSize(queue.getTaskRuntime(taskId)));
} catch (Exception e) {
  LOG.warn("Failed to track compaction started for task {}", taskId, e);
}
```

amoro-ams/src/test/java/org/apache/amoro/server/utils/KerberizedTestHelper.java
badalprasadsingh, line 1: Is this a test file? Was it discussed to add a test file? If not, please remove it.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+    return new LocationInfo("NA", "NA", "NA");
+  }
+
+  private PlatformInfo gatherPlatformInfo() {
+    String os = System.getProperty("os.name", "Unknown").toLowerCase();
+    String arch = System.getProperty("os.arch", "Unknown");
+    int cores = Runtime.getRuntime().availableProcessors();
+
+    String driverVersion = System.getenv("DRIVER_VERSION");
```
badalprasadsingh: Why do we need `DRIVER_VERSION`? In case it is `OLake` specific, please remove it. In OLake-Fusion, _driver_ means _optimizer_

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,259 @@
...
+    if (driverVersion == null || driverVersion.isEmpty()) {
+      driverVersion = "Not Available";
+    }
+
+    return new PlatformInfo(os, arch, driverVersion, cores + " cores");
+  }
+
+  private String resolveUserID() {
+    String configPathStr = System.getProperty("amoro.config.path", System.getProperty("user.home"));
```
badalprasadsingh: Where do we set `amoro.config.path` in the codebase? If we don't set it anywhere, it will always fall back to `user.home`.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,259 @@
...
+    this.objectMapper = new ObjectMapper();
+
+    initAsync();
+  }
+
+  private void initAsync() {
+    CompletableFuture.runAsync(
+        () -> {
+          try {
+            String disabledEnv = System.getenv("TELEMETRY_DISABLED");
```
badalprasadsingh: Have we tested it with `TELEMETRY_DISABLED`? If not, we have to. Its important.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,259 @@
...
+  private static final class InstanceHolder {
+    private static final Telemetry INSTANCE = new Telemetry();
+  }
+
+  public static Telemetry getInstance() {
+    return InstanceHolder.INSTANCE;
+  }
+
+  private Telemetry() {
+    this.httpClient = HttpClient.newBuilder().connectTimeout(Duration.ofSeconds(5)).build();
```
badalprasadsingh: Make the connection Timeout to `10min`. I know this is very high, but we have set it as default for every AMS API in OLake-UI

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,259 @@
...
+    String arch = System.getProperty("os.arch", "Unknown");
+    int cores = Runtime.getRuntime().availableProcessors();
+
+    String driverVersion = System.getenv("DRIVER_VERSION");
+    if (driverVersion == null || driverVersion.isEmpty()) {
+      driverVersion = "Not Available";
+    }
+
+    return new PlatformInfo(os, arch, driverVersion, cores + " cores");
+  }
```
badalprasadsingh, line 193: Here, we need to discuss few things with Product: 

1. First, `os` details are to be returned: if yes, this codebase will only return the `os` details of Fusion POD. Confirm: if we need to pass details of `optimizer` and `executor` as well.
2. Second: confirm if we should pass `cores` as well? In case yes, again, we have to pass `RAM memory` and `cores` and configuration params of `optimizer` and `executors` as well. Right now, we only return the `cores` of AMS pod.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,259 @@
...
+    Path idPath = Paths.get(configPathStr, USER_ID_FILE);
+
+    try {
+      if (Files.exists(idPath)) {
+        return Files.readString(idPath, StandardCharsets.UTF_8).trim().replace("\"", "");
+      }
+
+      MessageDigest digest = MessageDigest.getInstance("SHA-256");
+      String seed = String.valueOf(System.currentTimeMillis());
+      byte[] encodedHash = digest.digest(seed.getBytes(StandardCharsets.UTF_8));
```
badalprasadsingh: In OLake as well we create the `hash` the same way right?

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
badalprasadsingh: Make all connection timeouts in this PR to `10min`

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,259 @@
...
+            LOG.debug("Failed to initialize telemetry context safely: {}", e.getMessage());
+          }
+        });
+  }
+
+  private String fetchOutboundIP() {
+    try {
+      HttpRequest request =
+          HttpRequest.newBuilder()
+              .uri(URI.create("https://api.ipify.org?format=text"))
```
badalprasadsingh: nit: create all this `https:` constants at one place, and reuse it (`https://api.ipify.org?format=text`)

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  private LocationInfo fetchLocationFromIP(String ip) {
+    if (IP_NOT_FOUND_PLACEHOLDER.equals(ip)) {
+      return new LocationInfo("NA", "NA", "NA");
+    }
+    try {
+      HttpRequest request =
+          HttpRequest.newBuilder()
+              .uri(URI.create("https://ipinfo.io/" + ip + "/json"))
```
badalprasadsingh: Make this constant as well.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,259 @@
...
+    } catch (IOException | NoSuchAlgorithmException e) {
+      LOG.debug("Failed to parse or save telemetry user id file, using transient identity.", e);
+      return "transient-" + java.util.UUID.randomUUID().toString().substring(0, 8);
+    }
+  }
+
+  private void sendEvent(String eventName, Map<String, Object> props) {
+    if (platform == null) {
+      return; // Engine initialization was skipped or telemetry is intentionally disabled
+    }
```
badalprasadsingh: nit: Instead of checking just on the platform variable, create a dedicated function which tells if Telemetry is Disabled/Enabled and use that as the `single-source-of-truth` for all these logical areas.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  private static final double BYTES_PER_GB = 1024.0 * 1024.0 * 1024.0;
+
+  private static double bytesToGb(long bytes) {
+    return bytes / BYTES_PER_GB;
+  }
+
+  public void trackCompactionStarted(OptimizingType compactionType, long tableSize) {
```
badalprasadsingh: Where in this PR we are mentioning `Compaction` replace it with `Optimization`

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,259 @@
...
+    Map<String, Object> props =
+        Map.of("Optimization_Type", compactionType.name(), "table_size", bytesToGb(tableSize));
```
badalprasadsingh: Here is a problem.

Every time we return the type: from OLake-Fusion we return `MINOR` but from OLake-UI we will return `LITE`. It should be consistent.

Create a helper function which switches based on `MINOR` or `MAJOR` and returns `LITE` or `MEDIUM` as the compaction type.

Use that helper function everywhere wherever we are returning compaction types.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,259 @@
...
+      Map<String, Object> enrichedProperties = new HashMap<>(props);
+      enrichedProperties.put("OS", platform.os());
+      enrichedProperties.put("Arch", platform.arch());
+      enrichedProperties.put("Olake-Fusion_Version", platform.version());
+      enrichedProperties.put("Num_CPU", platform.deviceCpu());
+      enrichedProperties.put("IP_Address", ipAddress);
+      enrichedProperties.put("Location", locationInfo);
+      enrichedProperties.put("Distinct_ID", userID);
+      enrichedProperties.put("Time", System.currentTimeMillis() / 1000L);
+      enrichedProperties.put("Event_Original_Name", eventName);
```
badalprasadsingh: We need to make it the same case for the telemetry metrics as it is in `OLake`. Follow the same case as we have there, as far as I remember its `snake_case`.

As Product will run queries on the top of these metrics different naming convention becomes a problem.

badalprasadsingh (changes requested):
Second Review. Will test after this.

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
ashisharma-dz: I think since completetask() is called after executetask where the error in case of a failure is already being sent to the optimizer, completetask already has a taskresult with success or failure, so do we need to send it again to the optimizer?

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
ashisharma-dz: Have let it be set to user.home only for now

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
```diff
@@ -181,6 +182,7 @@ public void createCatalog(CatalogMeta catalogMeta) {
       throw new AlreadyExistsException("Catalog " + catalogMeta.getCatalogName());
     }
     fillCatalogProperties(catalogMeta);
+    Telemetry.getInstance().trackCatalogAdded(catalogMeta.getCatalogType());
```
badalprasadsingh: One more question: on `UI`, we deliver the metric of catalog creation clicked and button pressed etc., even though the catalog is not successfully created, right?

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -228,7 +232,15 @@ public OptimizingTask pollTask(String authToken, int threadId) {
...
-    queue.ackTask(taskId, getAuthenticatedOptimizer(authToken).getThread(threadId));
+    try {
+      OptimizingType optimizingType = queue.getOptimizingType(taskId);
+      TaskRuntime<?> taskRuntime = queue.getTaskRuntime(taskId);
+      queue.ackTask(taskId, getAuthenticatedOptimizer(authToken).getThread(threadId));
+      Telemetry.getInstance()
+          .trackOptimizationStarted(optimizingType, getInputTableSize(taskRuntime));
+    } catch (Exception e) {
+      LOG.warn("Failed to track optimization started for task {}", taskId, e);
+    }
```
badalprasadsingh: Same goes for this.

We are passing the Optimization or Compaction started metric when the sub-task is acknowledged. This is not correct.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
badalprasadsingh: Make a constant of `10min` and use it everywhere.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,267 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+public class Telemetry {
```
badalprasadsingh, line 51: Add a `TODO` for the Spark metrics for now.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,267 @@
...
+          }
+        });
+  }
+
+  private String fetchOutboundIP() {
+    try {
+      HttpRequest request =
+          HttpRequest.newBuilder()
+              .uri(URI.create(IPIFY_URL))
+              .timeout(Duration.ofSeconds(5))
```
badalprasadsingh: Use the constant of `10min` at all these places like this.

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -240,9 +252,29 @@ public void completeTask(String authToken, OptimizingTaskResult taskResult) {
...
+      TaskRuntime<?> taskRuntime = queue.getTaskRuntime(taskId);
+      Telemetry.getInstance()
+          .trackOptimizationCompleted(optimizingType, getInputTableSize(taskRuntime), success);
+      OptimizerThread thread =
+          getAuthenticatedOptimizer(authToken).getThread(taskResult.getThreadId());
+      queue.completeTask(thread, taskResult);
+    } catch (Exception e) {
+      LOG.warn("Failed to track optimization completed for task {}", taskId, e);
+    }
+  }
```
badalprasadsingh: Need to remove this. 

In Fusion, when any optimization (or compaction) runs. That is a single process, and that process generates sub-tasks based on its requirements and runs it across Spark Executors.

Right now the `trackOptimizationCompleted` metric is sent when a sub-task is marked completed (not the compaction as a process) we need to remove it completely from here.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
badalprasadsingh, line 790: We should be adding the `trackOptimizationCompleted` here, as this is the place where the optimization is finally completed given `status` == `ProcessStatus.SUCCESS`

badalprasadsingh, line 915: We need to add the `trackOptimizationStarted` metric here. This is the actual area where the compaction process starts. Subsequently the `inputTableSize` should come right below it.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -368,6 +368,14 @@ private TableOptimizingProcess planInternal(DefaultTableRuntime tableRuntime) {
...
   }
 
+  public TaskRuntime<?> getTaskRuntime(OptimizingTaskId taskId) {
+    return findProcess(taskId).getTaskRuntime(taskId);
+  }
+
+  public OptimizingType getOptimizingType(OptimizingTaskId taskId) {
+    return findProcess(taskId).getOptimizingType();
+  }
+
```
badalprasadsingh: We should ultimately remove this as we would no longer be dealing with sub-tasks.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,267 @@
...
+
+  private PlatformInfo gatherPlatformInfo() {
+    String os = System.getProperty("os.name", "Unknown").toLowerCase();
+    String arch = System.getProperty("os.arch", "Unknown");
+    int cores = Runtime.getRuntime().availableProcessors();
+
+    return new PlatformInfo(os, arch, cores + " cores");
+  }
+
+  private String resolveUserID() {
```
badalprasadsingh, line 178: Here is the problem.

When `olake-ui` tech stack is run completely. We create two different `user-ids` one for OLake-Go and one for Fusion. Those two are different as it is based on `time.now()`.

This should not happen, as we won't be able to correlate the user then.

So, we need to take the user-id from where OLake-UI stores it (`/tmp/olake-config/telemetry/user_id`).

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,267 @@
...
+          "table_size",
+          bytesToGb(tableSize));
+    sendEvent("Optimization Started - Fusion", props);
+  }
+
+  public void trackOptimizationCompleted(
+      OptimizingType optimizationType, long tableSize, boolean success) {
+    Map<String, Object> props =
+        Map.of(
+            "Optimization_Type",
```
badalprasadsingh: Make it `optimization_type`

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,267 @@
...
+            optimizationTypeHelper(optimizationType.name()),
+            "table_size",
+            bytesToGb(tableSize),
+            "Optimization_Status",
+            success ? "SUCCESS" : "FAILED");
+    sendEvent("Optimization Completed - Fusion", props);
+  }
+
+  public void trackCatalogAdded(String catalogType) {
+    Map<String, Object> props = Map.of("Catalog_Type", catalogType);
```
badalprasadsingh: Make it `catalog_type`

badalprasadsingh (changes requested):
Final Review.

Please run `mvn spotless:apply`, that should resolve the CI failure.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
badalprasadsingh: Bring it down just before the log line.

```java
  public void createCatalog(CatalogMeta catalogMeta) {
    if (catalogExist(catalogMeta.getCatalogName())) {
      throw new AlreadyExistsException("Catalog " + catalogMeta.getCatalogName());
    }
    fillCatalogProperties(catalogMeta);
    validateCatalogConnection(catalogMeta);
    // Build to make sure the catalog is valid
    ServerCatalog catalog = CatalogBuilder.buildServerCatalog(catalogMeta, serverConfiguration);
    doAs(CatalogMetaMapper.class, mapper -> mapper.insertCatalog(catalog.getMetadata()));
    disposeCatalog(catalogMeta.getCatalogName());
    serverCatalogMap.put(catalogMeta.getCatalogName(), catalog);
    Telemetry.getInstance().trackCatalogAdded(catalogMeta.getCatalogType());
    LOG.info(
        "Create catalog {}, type:{}", catalogMeta.getCatalogName(), catalogMeta.getCatalogType());
  }
```

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -228,7 +228,10 @@ public OptimizingTask pollTask(String authToken, int threadId) {
   public void ackTask(String authToken, int threadId, OptimizingTaskId taskId) {
     LOG.info("Ack task {} by optimizer {} (threadId {})", taskId, authToken, threadId);
     OptimizingQueue queue = getQueueByToken(authToken);
-    queue.ackTask(taskId, getAuthenticatedOptimizer(authToken).getThread(threadId));
+    try {
+      queue.ackTask(taskId, getAuthenticatedOptimizer(authToken).getThread(threadId));
+    } catch (Exception e) {
+    }
```
badalprasadsingh: Remove these changes. These are unnecessary.

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -240,9 +243,14 @@ public void completeTask(String authToken, OptimizingTaskResult taskResult) {
...
-        getAuthenticatedOptimizer(authToken).getThread(taskResult.getThreadId());
-    queue.completeTask(thread, taskResult);
+    OptimizingTaskId taskId = taskResult.getTaskId();
+    boolean success = taskResult.getErrorMessage() == null;
+    try {
+      OptimizerThread thread =
+          getAuthenticatedOptimizer(authToken).getThread(taskResult.getThreadId());
+      queue.completeTask(thread, taskResult);
+    } catch (Exception e) {
+    }
```
badalprasadsingh: Remove these changes. These are unnecessary now.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -777,9 +778,13 @@ public void commit() {
           buildCommit().commit();
           if (allTasksPrepared()) {
             status = ProcessStatus.SUCCESS;
+            Telemetry.getInstance()
+                .trackOptimizationCompleted(optimizingType, tableRuntime.getTableSize(), true);
```
badalprasadsingh: For all these places create a helper function like this, and reuse it as `trackCompleted(false)` or `trackCompleted(true)` whatsoever.

```java
    private void trackCompleted(boolean success) {
      try {
        Telemetry.getInstance()
            .trackOptimizationCompleted(optimizingType, inputTableSize(), success);
      } catch (Exception e) {
        LOG.debug("Failed to track optimization completed for process {}", processId, e);
      }
    }
```

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -777,9 +778,13 @@ public void commit() {
...
           } else if (taskMap.values().stream()
               .anyMatch(task -> task.getStatus() == TaskRuntime.Status.FAILED)) {
             status = ProcessStatus.FAILED;
+            Telemetry.getInstance()
```
badalprasadsingh: Use this `trackCompleted` function at all these sites.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
badalprasadsingh, line 569: If the user forcefully closes/cancels the compaction, we need to send the telemetry metric then as well.

Please call `trackCompleted(false)`

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -882,6 +889,7 @@ private void beginAndPersistProcess() {
                   mapper -> mapper.insertTaskRuntimes(Lists.newArrayList(taskMap.values()))),
           () -> TaskFilesPersistence.persistTaskInputs(processId, taskMap.values()),
           () -> tableRuntime.beginProcess(this));
+      Telemetry.getInstance().trackOptimizationStarted(optimizingType, tableRuntime.getTableSize());
```
badalprasadsingh: Use the `inputTableSize` instead of `tableRuntime.getTableSize()` function we had earlier.

```java
    private long inputTableSize() {
      try {
        MetricsSummary summary = getSummary();
        if (summary != null && summary.getInputFilesStatistics() != null) {
          return summary.getInputFilesStatistics().getTotalSize();
        }
      } catch (Exception e) {
        LOG.debug("Failed to compute input table size for telemetry, process {}", processId, e);
      }
      return 0;
    }
```

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,268 @@
...
+//  Desired Parallelism)
+
+public class Telemetry {
+  private static final Logger LOG = LoggerFactory.getLogger(Telemetry.class);
+
+  private static final String TRACK_URL = "https://analytics.olake.io/mp/track";
+  private static final String IPINFO_URL = "https://ipinfo.io/";
+  private static final String IPIFY_URL = "https://api.ipify.org?format=text";
+  private static final String IP_NOT_FOUND_PLACEHOLDER = "NA";
+  private static final String USER_ID_FILE = "user_id.txt";
```
badalprasadsingh: This variable `USER_ID_FILE` is not used anymore remove this line.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,268 @@
...
+  private static final String USER_ID_FILE_ENV = "USER_ID_FILE_ENV";
+
+  private static final long TIMEOUT_MINS = 10L;
+
+  private final HttpClient httpClient;
+  private final ObjectMapper objectMapper;
+
+  private String ipAddress = IP_NOT_FOUND_PLACEHOLDER;
+  private PlatformInfo platform;
+  private LocationInfo locationInfo;
```
badalprasadsingh: We need to add the `userID` variable. We missed it. `private volatile String userID;`

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,268 @@
...
+
+  public void trackOptimizationCompleted(
+      OptimizingType optimizationType, long tableSize, boolean success) {
+    Map<String, Object> props =
+        Map.of(
+            "optimization_type",
+            optimizationTypeHelper(optimizationType.name()),
+            "table_size",
+            bytesToGb(tableSize),
+            "Optimization_Status",
```
badalprasadsingh: Use `optimization_status`

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    try {
+      Map<String, Object> enrichedProperties = new HashMap<>(props);
+      enrichedProperties.put("os", platform.os());
+      enrichedProperties.put("arch", platform.arch());
+      enrichedProperties.put("num_cpu", platform.deviceCpu());
+      enrichedProperties.put("ip_address", ipAddress);
+      enrichedProperties.put("location", locationInfo);
+      enrichedProperties.put("distinct_id", userID);
```
badalprasadsingh: Fetch it when its null : `enrichedProperties.put("distinct_id", userID != null ? userID : resolveUserID())`

badalprasadsingh (changes requested):
Mostly approved all minor comments.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,268 @@
...
+    }
+  }
+
+  private static final double BYTES_PER_GB = 1024.0 * 1024.0 * 1024.0;
+
+  private static double bytesToGb(long bytes) {
+    return bytes / BYTES_PER_GB;
+  }
+
+  public String optimizationTypeHelper(String optimizationTypeName) {
```
badalprasadsingh: non-blocking: make it `private` please

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,268 @@
...
+    return new PlatformInfo(os, arch, cores + " cores");
+  }
+
+  private String resolveUserID() {
+    String configuredPath = System.getenv(USER_ID_FILE_ENV);
+    Path shared =
+        Paths.get(
+            (configuredPath != null && !configuredPath.isBlank())
+                ? configuredPath.trim()
+                : SHARED_USER_ID_FILE);
```
badalprasadsingh: Let's only read from the shared mounted dir. Remove `USER_ID_FILE_ENV`

```java
  private String resolveUserID() {
    Path shared = Paths.get(SHARED_USER_ID_FILE);
    try {
      if (Files.exists(shared)) {
        String id = Files.readString(shared, StandardCharsets.UTF_8).trim().replace("\"", "");
        if (!id.isEmpty()) {
          return id;
        }
      }
    } catch (IOException e) {
      LOG.debug("Failed to read shared telemetry user id at {}: {}", shared, e.getMessage());
    }
    return null;
  }
```

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,268 @@
...
+public class Telemetry {
+  private static final Logger LOG = LoggerFactory.getLogger(Telemetry.class);
+
+  private static final String TRACK_URL = "https://analytics.olake.io/mp/track";
+  private static final String IPINFO_URL = "https://ipinfo.io/";
+  private static final String IPIFY_URL = "https://api.ipify.org?format=text";
+  private static final String IP_NOT_FOUND_PLACEHOLDER = "NA";
+  private static final String USER_ID_FILE = "user_id.txt";
+  private static final String SHARED_USER_ID_FILE = "/tmp/olake-config/telemetry/user_id";
+  private static final String USER_ID_FILE_ENV = "USER_ID_FILE_ENV";
```
badalprasadsingh: Remove this line as we no longer need it, `USER_ID_FILE_ENV`

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -882,6 +896,19 @@ private void beginAndPersistProcess() {
                   mapper -> mapper.insertTaskRuntimes(Lists.newArrayList(taskMap.values()))),
           () -> TaskFilesPersistence.persistTaskInputs(processId, taskMap.values()),
           () -> tableRuntime.beginProcess(this));
+      Telemetry.getInstance().trackOptimizationStarted(optimizingType, getInputTableSize());
```
badalprasadsingh: The function we have is `inputTableSize()` not `getInputTableSize()`

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
badalprasadsingh, line 665: Need to add ` trackCompleted(false);` here.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -777,9 +779,11 @@ public void commit() {
           buildCommit().commit();
           if (allTasksPrepared()) {
             status = ProcessStatus.SUCCESS;
+            trackCompleted(true);
           } else if (taskMap.values().stream()
               .anyMatch(task -> task.getStatus() == TaskRuntime.Status.FAILED)) {
             status = ProcessStatus.FAILED;
+            trackCompleted(false);
           } else {
             status = ProcessStatus.CLOSED;
```
badalprasadsingh, line 788: Need to add `trackCompleted(false);` here as well.

badalprasadsingh (changes requested):

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -240,6 +240,8 @@ public void completeTask(String authToken, OptimizingTaskResult taskResult) {
         taskResult.getTaskId(),
         taskResult.getErrorMessage() == null ? "SUCCESS" : "FAIL");
     OptimizingQueue queue = getQueueByToken(authToken);
+    OptimizingTaskId taskId = taskResult.getTaskId();
+    boolean success = taskResult.getErrorMessage() == null;
```
badalprasadsingh: Dead Code. Please remove it.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -777,11 +780,14 @@ public void commit() {
...
           } else if (taskMap.values().stream()
               .anyMatch(task -> task.getStatus() == TaskRuntime.Status.FAILED)) {
             status = ProcessStatus.FAILED;
+            trackCompleted();
           } else {
             status = ProcessStatus.CLOSED;
+            trackCompleted();
           }
           endTime = System.currentTimeMillis();
           persistAndSetCompleted(status == ProcessStatus.SUCCESS);
```
badalprasadsingh, line 791: There is a possibility of double emission here, please make it like this:

```java
        boolean tracked = false;
        try {
          hasCommitted = true;
          buildCommit().commit();
          if (allTasksPrepared()) {
            status = ProcessStatus.SUCCESS;
          } else if (taskMap.values().stream()
              .anyMatch(task -> task.getStatus() == TaskRuntime.Status.FAILED)) {
            status = ProcessStatus.FAILED;
          } else {
            status = ProcessStatus.CLOSED;
          }
          trackCompleted();
          tracked = true;
          endTime = System.currentTimeMillis();
          persistAndSetCompleted(status == ProcessStatus.SUCCESS);
```

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -793,6 +799,7 @@ public void commit() {
         } catch (Throwable t) {
           LOG.error("{} Commit optimizing failed ", tableRuntime.getTableIdentifier(), t);
           status = ProcessStatus.FAILED;
+          trackCompleted();
```
badalprasadsingh:
```java
          if (!tracked) {
            trackCompleted();
          }
```

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -882,6 +909,19 @@ private void beginAndPersistProcess() {
                   mapper -> mapper.insertTaskRuntimes(Lists.newArrayList(taskMap.values()))),
           () -> TaskFilesPersistence.persistTaskInputs(processId, taskMap.values()),
           () -> tableRuntime.beginProcess(this));
+      Telemetry.getInstance().trackOptimizationStarted(optimizingType, inputTableSize());
```
badalprasadsingh: Minor: We don't add it in a try catch block here. Please do it. Function `trackStarted`

```java
      trackStarted();
    }

    private void trackStarted() {
      try {
        Telemetry.getInstance().trackOptimizationStarted(optimizingType, inputTableSize());
      } catch (Exception e) {
        LOG.debug("Failed to track optimization started for process {}", processId, e);
      }
```

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    try {
+      Map<String, Object> enrichedProperties = new HashMap<>(props);
+      enrichedProperties.put("os", platform.os());
+      enrichedProperties.put("arch", platform.arch());
+      enrichedProperties.put("num_cpu", platform.deviceCpu());
+      enrichedProperties.put("ip_address", ipAddress);
+      enrichedProperties.put("location", locationInfo);
+      enrichedProperties.put("distinct_id", resolveUserID());
```
badalprasadsingh: Minor. Take it from the cache or read again from the file.

```java
enrichedProperties.put("distinct_id", userID != null ? userID : resolveUserID());
```

badalprasadsingh (changes requested):

badalprasadsingh (changes requested):
We need to raise a PR on `olake-helm` to mount the path of `/tmp/olake-config/telemetry` for the Fusion Container.

Something like this: 
```yaml
            - name: shared-storage
              mountPath: /tmp/olake-config/telemetry
              subPath: telemetry
              readOnly: true
```

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,308 @@
...
+public class Telemetry {
+  private static final Logger LOG = LoggerFactory.getLogger(Telemetry.class);
+
+  private static final String TRACK_URL = "https://analytics.olake.io/mp/track";
+  private static final String IPINFO_URL = "https://ipinfo.io/";
+  private static final String IPIFY_URL = "https://api.ipify.org?format=text";
+  private static final String IP_NOT_FOUND_PLACEHOLDER = "NA";
+  private static final String SHARED_USER_ID_FILE = "/tmp/olake-config/telemetry/user_id";
+
+  private static final long TIMEOUT_MINS = 10L;
```
badalprasadsingh: Lets change it from `10mins` to what ever limit we are setting in UI.

And, change these lines of code to use seconds in place of minutes where ever used:
```java
HttpClient.newBuilder().connectTimeout(Duration.ofMinutes(TIMEOUT_MINS)).build();
    this.objectMapper = new ObjectMapper();
```

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizerGroupMetrics.java
```diff
@@ -14,6 +14,8 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
+ * Modified by Datazip Inc. in 2026
```
badalprasadsingh: We are not changing this file anymore in this PR. We can remove this in this PR.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,318 @@
...
+import java.security.NoSuchAlgorithmException;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+
+// TODO: Add spark metrics  (Spark Driver Memory, Spark Executor Memory, Spark Executor Cores,
+//  Desired Parallelism)
```
badalprasadsingh: Need to remove this TODO now

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,318 @@
...
+      LocationInfo loc = locationInfo != null ? locationInfo : new LocationInfo("NA", "NA", "NA");
+      Map<String, Object> enrichedProperties = new HashMap<>(props);
+      enrichedProperties.put("os", p.os());
+      enrichedProperties.put("arch", p.arch());
+      enrichedProperties.put("num_cpu", p.deviceCpu());
+      enrichedProperties.put("ip_address", ipAddress);
+      enrichedProperties.put("location", loc);
+      enrichedProperties.put("distinct_id", userID != null ? userID : resolveUserID());
+      enrichedProperties.put("time", System.currentTimeMillis() / 1000L);
+      enrichedProperties.put("event_original_name", eventName);
```
badalprasadsingh, line 266: Lets remove `num_cpu`, `arch` as we are sending for Spark now.

These would be of Fusion Pod, but it will be sent on every event which is not good.

amoro-ams/src/main/java/org/apache/amoro/server/utils/Telemetry.java
```diff
@@ -0,0 +1,318 @@
...
+    return new LocationInfo("NA", "NA", "NA");
+  }
+
+  private PlatformInfo gatherPlatformInfo() {
+    String os = System.getProperty("os.name", "Unknown").toLowerCase();
+    String arch = System.getProperty("os.arch", "Unknown");
+    int cores = Runtime.getRuntime().availableProcessors();
+
+    return new PlatformInfo(os, arch, cores + " cores");
+  }
```
badalprasadsingh, line 176: We just have to pass `os.name` now, others we can remove. These are of Fusion POD.

badalprasadsingh (changes requested):

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -796,12 +801,38 @@ public void commit() {
...
+      try {
+        Telemetry.getInstance()
+            .trackOptimizationCompleted(
+                optimizingType,
+                inputTableSize(),
+                telemetryStatus(),
+                getDuration(),
+                quotaProvider.getTotalQuota(optimizerGroup.getName()));
+      } catch (Exception e) {
+        LOG.debug("Failed to track optimization completed for process {}", processId, e);
```
badalprasadsingh: Make it `Throwable`.

```java
      } catch (Throwable t) {
        LOG.debug("Failed to track optimization completed for process {}", processId, t);
```

hash-data:
This is AI review we can check it 
```
Blockers — telemetry DOES disturb fusion behavior

1. DefaultCatalogManager.updateCatalog now swallows every exception. DefaultCatalogManager.java:246-262 — new catch (Exception e) { track(false); LOG.error(...) } with no rethrow. CatalogController.java:697 calls it then unconditionally ctx.json(OkResponse.ok()). Result: invalid catalog type change, bad connection, non-existent catalog → API returns 200 OK, UI shows success, nothing updated. Real regression, plus it breaks existing test TestCatalogService.testUpdateCatalog:118-124 which asserts IllegalMetadataException and ObjectNotExistsException propagate.
Fix: track failure, then throw.

2. Telemetry calls in catalog paths are unguarded. Every other call site wraps in try/catch Throwable (OptimizingQueue.trackStarted/trackCompleted, DefaultOptimizingService.trackInstalledFusion). Catalog paths do not. In createCatalog, if trackCatalogCreated(...,true) on the success path throws, control falls into catch → emits a success=false event → rethrows RuntimeException even though the catalog was committed and cached. Also, the Telemetry call inside catch can throw and destroy the original exception.
Fix: wrap both call sites same as the others.

3. trackInstalledFusion fires per optimizer, not per install. DefaultOptimizingService.java:168 — loadOptimizingQueues calls registerOptimizer in a loop at boot, and authenticate calls it on every optimizer (re)registration. Each call runs a full OptimizerMapper.selectAll and emits one "Installed Fusion" event. N optimizers = N events + N table scans at every AMS start.
Fix: fire once after loadOptimizingQueues, or debounce by token.

Should fix

4. Double getSummary() under the process lock. OptimizingQueue.java:927 — inputTableSize() rebuilds MetricsSummary over all tasks, while persistAndSetCompleted already builds one; both run inside lock.lock() in close()/commit(). Compute once and pass it in.

5. Blocking HTTP on ForkJoinPool.commonPool. Telemetry.initAsync → CompletableFuture.runAsync does two synchronous httpClient.send calls (ipify + ipinfo, 10s timeout each) on the shared JVM pool; sendEvent also chains thenRunAsync there. On a 1-2 core container that squats a commonPool thread up to 20s at startup. Use a dedicated daemon single-thread executor.

6. distinct_id not durable on k8s. SHARED_USER_ID_FILE = /tmp/olake-config/telemetry/user_id. Only local-test/docker-compose.yml mounts it. Helm deploys (the branch reports deployment_mode=HELM) get a fresh id every pod restart → inflated unique-install counts.

7. Opt-out is env-var only and undocumented. TELEMETRY_DISABLED appears nowhere but Telemetry.java:106 — no config.yaml key, no README/chart value. Add a config option and document the three outbound endpoints (analytics.olake.io, api.ipify.org, ipinfo.io).

Minor

- createCatalog's e instanceof RuntimeException ? … : new RuntimeException(e) is dead — the method declares no checked exceptions, so nothing checked can reach it. Drop the wrap.
- AlreadyExistsException (pre-existing catalog) now emits Catalog Created success=false. Confirm that's wanted, or exclude it.
- Commit path sets ProcessStatus.CLOSED for a partial commit; telemetryStatus() reports it as CANCELLED. OptimizingQueue.java:826-833.
- resolveSparkConfig returns an empty map when the container has no properties, so spark_* keys vanish instead of being "NA". Inconsistent with the populated path.
- Multiple Spark containers → last one wins in AmoroServiceContainer.java:641.
- getSparkConfig() hands out the mutable HashMap. Wrap in Map.copyOf / unmodifiableMap.
- .gitignore /.cocoindex_code/ entry is uncommitted local work, not in the PR — decide if it ships.
```

PR #16, opened 2026-07-20
fix: fusion corrupted tables

ashisharma-dz:
## Brief change log

- Cooldown Tracking: Added a thread-safe map (planFailureCooldownUntil) and a 60-second time constant to OptimizingQueue.java to temporarily track table identifiers that fail their planning stage.

- Modified the .whenComplete handler inside triggerAsyncPlanning to catch exceptions during table planning, automatically registering the failed table into the cooldown map to halt tight-loop re-scheduling.

- Scheduler: Updated scheduleTableIfNecessary with a new helper method (skipCooldownTables) that dynamically purges expired cooling periods and appends active cooling tables to the existing skipTables set.

## How was this patch tested?

- Live tested by reproducing the issue, adding 2 tables to one catalog, corrupting the metadata for one table and triggering compaction
- After the fix: the corrupted table compaction fails (status goes back to idle), and similarly the healthy table successfully transitions from idle->pending->planning->optimizing->SUCCESS ->idle

## Documentation

- Does this pull request introduce a new feature? (yes / no) no
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) not applicable


amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -110,6 +110,9 @@ public class OptimizingQueue extends PersistentBase {
   private final Executor planExecutor;
   // Keep all planning table identifiers
   private final Set<ServerTableIdentifier> planningTables = new HashSet<>();
+  private final Map<ServerTableIdentifier, Long> planFailureCooldownUntil =
+      new ConcurrentHashMap<>();
+  private static final long PLAN_FAILURE_COOLDOWN_MS = 60_000L; // current: 1 min, tune if needed
```
badalprasadsingh: Right now, the problem is: every iceberg table which is _coruppted_, is retried after every `1 min`, which should not happen. As think of if 200 iceberg tables are corrupted all those would be retried in a span of `60s`.

Better: retry the corrupted iceberg table for compaction only when its `cron` time comes.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -368,6 +384,14 @@ private TableOptimizingProcess planInternal(DefaultTableRuntime tableRuntime) {
...
   }
 
+  public TaskRuntime<?> getTaskRuntime(OptimizingTaskId taskId) {
+    return findProcess(taskId).getTaskRuntime(taskId);
+  }
+
+  public OptimizingType getOptimizingType(OptimizingTaskId taskId) {
+    return findProcess(taskId).getOptimizingType();
+  }
+
```
badalprasadsingh: Are we even using these `getters` anywhere? In case not, please remove them.

badalprasadsingh (changes requested):
Given First Review.

We are missing here on the main issue: An corrupted iceberg table which is not enabled for compaction ( `self-optimizing.enabled` = `true/false` ) should not block the compaction phase of any iceberg table, which is fine.

Other than this, please fix these:

1. Please change the base branch for the PR to `staging` from master.
2. A better PR Title should start with `fix:` instead of `Fix/`.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -368,6 +369,10 @@ private TableOptimizingProcess planInternal(DefaultTableRuntime tableRuntime) {
     }
   }
 
+  public TaskRuntime<?> getTaskRuntime(OptimizingTaskId taskId) {
+    return findProcess(taskId).getTaskRuntime(taskId);
+  }
+
```
badalprasadsingh: This function is unnecessary right? Please remove it.

badalprasadsingh (changes requested):
Second Review. Will test after this.

Please change the base branch to `staging` from `master`.

badalprasadsingh:
MAJOR: Right now, this only fixes the scenario where the _corrupted_ iceberg table is enabled (`self.optimizing-enabled=true`) for compaction. It does not fixes the scenario where the corrupted iceberg table is **not** enabled for compaction which is the most ideal case.

badalprasadsingh (dismissed):
This PR solve this part:

If an iceberg table is _enabled_ for compaction, and it continuously fails and continuously gets reset to PENDING state, it **blocks** the compaction of other iceberg tables.

Throwing it to IDLE state, lets Fusion handle this corrupted iceberg table in the next `cron` iteration, and thus breaking the blocked loop.

badalprasadsingh (approved):
Great. 

So, the issue was: one corrupted `database` stopped the whole `exploreExternalCatalog`. This fix is correct for that.

We are not logging out all the iceberg tables @hash-data, as we skip the whole database, not individual iceberg tables in the database. Reason: the issue is at database level. Thus, logging out each and every iceberg table in the database is redundant.

One consequence to take care of (not tested):
- The corrupted database might not appear on UI

PR #20, opened 2026-08-18
fix: cache snapshots

ashisharma-dz:
When the number of equality delete files is less than 2, the minor optimization is not considered necessary and skipped, even when the cron is fired.
With this change, in a case where the cron for minor has been fired and there is at least one delete file, the optimization will be marked as necessary and proceed for optimization

## Brief change log
- CommonPartitionEvaluator: isMinorNecessary() 
- A change of value from 1 to 0 in the return expression (bool)

## How was this patch tested?

## Documentation

- Does this pull request introduce a new feature? (yes / no) : NO
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented): not applicable

### DIFF
--- a/amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/CommonPartitionEvaluator.java
+++ b/amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/CommonPartitionEvaluator.java
@@ -409,8 +409,12 @@ public class CommonPartitionEvaluator implements PartitionEvaluator {
 
   public boolean isMinorNecessary() {
     int smallFileCount = fragmentFileCount + equalityDeleteFileCount;
+    LOG.warn("ASHI: isMinorNecessary partition={} fragmentFileCount={} equalityDeleteFileCount={} smallFileCount={} minorLeastFileCount={} reachMinorInterval={} combinePosSegmentFileCount={}",
+    getPartition(), fragmentFileCount, equalityDeleteFileCount,
+    fragmentFileCount + equalityDeleteFileCount, config.getMinorLeastFileCount(),
+    reachMinorInterval(), combinePosSegmentFileCount);
     return smallFileCount >= config.getMinorLeastFileCount()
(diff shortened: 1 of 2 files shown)


amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/CommonPartitionEvaluator.java
```diff
@@ -410,7 +410,7 @@ public boolean isMajorNecessary() {
   public boolean isMinorNecessary() {
     int smallFileCount = fragmentFileCount + equalityDeleteFileCount;
     return smallFileCount >= config.getMinorLeastFileCount()
-        || (smallFileCount > 1 && reachMinorInterval())
+        || (smallFileCount > 0 && reachMinorInterval())
```
badalprasadsingh, line 413: Here,
```
smallFileCount = fragmentFileCount + equalityDeleteFileCount
```

What this is currently solving is this case:
```
smallFileCount = 1
```
1. If only equality-detete file arrives but no fragment file (mostly, the soft-delete data file) -> trigger compaction
2. If only soft-delete data file arrives, but no equality delete file -> trigger compaction

Both these cases are unreached in an unpartitioned table. Because OLake always writes the corresponding soft-delete data file along with the data file.

Keep in mind the check is for a particular partition (whole table if it is un-partitioned).

In that case:
```
smallFileCount = 2
```

[IMPORTANT] Only case where this "might" be helpful is a partitioned table. (Not tested). Because if the partitioned value changes the delete file goes somewhere else and the soft-delete data file goes somewhere else. Thus for each partition `smallFileCount` can be `1`.

badalprasadsingh (dismissed):

hash-data:
The problem is even after 3 to 4 ingestion it is not triggering the compaction. that means the current fix is just a covering part actual problem is snapshot updating event when compaction not ran. Can we check this ?

hash-data (approved):

PR #26, opened 2026-09-01
hotfix(planner): plan compaction of all partitions in a single run

### LINKED ISSUE #29: [Bug]: All partitions in an Iceberg Table are not planned in an single optimization run by default
By default OLake-Fusion didn't plan to compact / optimize all partitions across an iceberg table in a single run. We need to fix this and by default all partitions must be planned in the run. 

End goal: There should not be any number of equality deletes left after any Lite / Medium / Full compaction. (Given: there is no ingestion continuously going on which adds equality-deletes past the snapshot from where compaction started)

badalprasadsingh:
Fixes #29, from now on, all partitions in an iceberg table will be "planned" for compaction in a single run.

Added a `TODO` to remove the cost calculation logics later on as removing it brings a lot of changes now.

### DIFF
--- a/.github/workflows/dev-image.yml
+++ b/.github/workflows/dev-image.yml
@@ -1,0 +1,40 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Modified by Datazip Inc. in 2026
+
+
+# Publishes the dev images on every push to feat/remove-partition-selection-logic.
+#
+# Because "dev-badal" does not match vX.X.X, docker-images.yml tags both images with
+# :dev-badal and :dev-badal-latest:
+# - olakego/fusion:dev-badal, olakego/fusion:dev-badal-latest
+# - olakego/fusion-spark:dev-badal, olakego/fusion-spark:dev-badal-latest
+
+name: Fusion Dev Image
+
+on:
+  push:
+    branches:
+      - feat/remove-partition-selection-logic-dev
+
+jobs:
+  build_fusion_dev:
+    name: Build and Publish Fusion Dev Image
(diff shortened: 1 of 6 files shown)


hash-data:
AI comments 
```
P0

1. rewrite_input BLOB still unbounded — real hard limit not fixed
OptimizingProcessMapper.java:74 updateProcessInputFiles writes all task inputs as ONE gzipped BLOB in one column (TaskFilesPersistence.java:39-46), and loadTaskInputs deserializes whole blob on recovery. Batching at OptimizingQueue.java:888 fixes only the 65535 bind-param limit, not this. All-partitions planning scales this blob with the table's total file count.
Failure: big table → serialize+gzip of every RewriteFilesInput in heap; MySQL max_allowed_packet (64MB default) rejects insert; whole doAsTransaction rolls back after all task_runtime batches already inserted in-tx, process fails, next plan produces same size → table wedged in permanent plan-fail loop, zero progress.
Fix: chunk rewrite_input per-task rows (or per-batch blobs) same way as task_runtime, or offload to object store.

2. No cap on tasks/memory per process
AbstractOptimizingPlanner.java:164 maxInputSize = Double.MAX_VALUE removes the only bound. splitTasks(targetTaskCount) ignores its arg in both splitters (AbstractPartitionPlan.java:372, MixedIcebergPartitionPlan.java:286) — task count = sum over all partitions of bytes/targetSize. All TaskRuntime + every data/delete file object live in taskMap simultaneously, get persisted, get fully re-loaded on AMS restart (OptimizingQueue.java:939). Prod already exceeded 4681 tasks per the new comment → AMS OOM is live risk, not hypothetical.
Fix: hard cap (max tasks or max files per process) with continuation on next round; the cap replaces maxInputSizePerThread, doesn't revert it.
                                                                                                                     
 3. optimizingType misclassification corrupts
AbstractOptimizingPlanner.java:172 adds every polled evaluator to actualPartitionPlans before splitTasks, so          partitions that produce zero tasks are includprocess type from that set — one FULL-necessary partition marks the whole process FULL. DefaultTableRuntime.java:448-454 sets only lastFullOptimizingTime;            lastMinorOptimizingTime / lastMajorOptimizing
Result: full cron recorded satisfied though most partitions got no full rewrite, minor/major cron conditions evaluate against stale timestamps forever. Pre-existins bounded before — now every process sees allpartitions, so it fires nearly always.
Fix: derive type from partitions that actualltype-specific timestamps from what was reallyrewritten.

```

badalprasadsingh:
1. This is for `MySQL`, we use `postgres`.
2. AI is asking to put a cap. If we put in any cap, all partitions won't fall into a single compaction process. That was our whole intention. 
3. This does not an issue. The thing is: during compaction even if the cron for the partitioned iceberg table is set to Full, a partition where Full requirements are not met, minor requirements are met, it will run minor. That's fine. As forcing Full will do the same thing. But AI says, to mark the last minor run to the partition's minor compaction timestamp. Which is wrong. The last minor run timestamp should be when the compaction process's cron is MINOR.

hash-data (approved):

PR #31, opened 2026-09-07
feat: plan fail as process fail

ashisharma-dz:
Today, if the planning itself for a table fails, it logs into ams but is never shown on the olake-UI as neither fail nor skip.
This PR marks such a process as fail, as well as appends the said ams error logs to the driver.

## Brief change log
- A process is inserted in beginPlanning() itself
- Said process marked as failed in planFailed
- In case the process goes ahead to actual optimization, the process to that process_id itself is updated
- The error message is appended to the driver

## How was this patch tested?

# Documentation

- [ ] Documentation Link: [link to README, olake.io/docs, or olake-docs]
- [x] N/A (bug fix, refactor, or test changes only)

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
@@ -137,7 +137,7 @@ public class OptimizingQueue extends PersistentBase {
           "Found orphaned PLANNING status for table {} on startup; marking process failed for re-scheduling.",
           tableRuntime.getTableIdentifier());
       tableRuntime.planFailed(
-          "orphaned PLANNING on AMS startup; process marked failed for re-schedule");
+          "orphaned PLANNING on AMS startup; process marked failed for re-schedule", null);
     }
 
     TableOptimizingProcess process = null;
@@ -346,7 +346,8 @@ public class OptimizingQueue extends PersistentBase {
               tableRuntime.refresh(table),
               (MixedTable) table.originalTable(),
               getAvailableCore(),
-              maxInputSizePerThread());
+              maxInputSizePerThread(),
+              tableRuntime.getProcessId());
       if (planner.isNecessary()) {
         return new TableOptimizingProcess(planner, tableRuntime);
       } else {
@@ -856,6 +857,14 @@ public class OptimizingQueue extends PersistentBase {
                           "",
                           new HashMap<>(),
                           getSummary().summaryAsMap(false))),
+          () ->
+              doAs(
+                  TableProcessMapper.class,
+                  mapper ->
+                      mapper.updateProcessType(
+                          tableRuntime.getTableIdentifier().getId(),
+                          processId,
+                          optimizingType.name().toUpperCase())),
           () ->
               doAs(
                   OptimizingProcessMapper.class,
--- a/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
@@ -57,7 +57,6 @@ import org.apache.amoro.shade.guava32.com.google.common.annotations.VisibleForTe
 import org.apache.amoro.shade.guava32.com.google.common.base.Preconditions;
 import org.apache.amoro.shade.guava32.com.google.common.collect.Lists;
 import org.apache.amoro.shade.guava32.com.google.common.collect.Maps;
-import org.apache.amoro.shade.jackson2.com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.amoro.table.MixedTable;
 import org.apache.amoro.table.TableIdentifier;
 import org.apache.amoro.utils.CompatiblePropertyUtil;
@@ -70,16 +69,8 @@ import org.apache.iceberg.util.StructLikeMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.nio.file.StandardOpenOption;
-import java.time.Instant;
-import java.time.ZoneOffset;
-import java.time.format.DateTimeFormatter;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -140,12 +131,13 @@ public class OptimizingQueue extends PersistentBase {
   }
 
   private void initTableRuntime(DefaultTableRuntime tableRuntime) {
-    // Recover tables stranded in PLANNING to PENDING
+    // Recover tables stranded in PLANNING: mark process FAILED then IDLE for re-scheduling
     if (tableRuntime.getOptimizingStatus() == OptimizingStatus.PLANNING) {
       LOG.warn(
-          "Found orphaned PLANNING status for table {} on startup; resetting to PENDING status for re-scheduling.",
+          "Found orphaned PLANNING status for table {} on startup; marking process failed for re-scheduling.",
           tableRuntime.getTableIdentifier());
-      tableRuntime.planFailed();
+      tableRuntime.planFailed(
+          "orphaned PLANNING on AMS startup; process marked failed for re-schedule");
     }
 
     TableOptimizingProcess process = null;
@@ -348,7 +340,6 @@ public class OptimizingQueue extends PersistentBase {
     tableRuntime.beginPlanning();
     try {
       ServerTableIdentifier identifier = tableRuntime.getTableIdentifier();
-      // mostly here corrupted metadata reacts:
       AmoroTable<?> table = catalogManager.loadTable(identifier.getIdentifier());
       AbstractOptimizingPlanner planner =
           IcebergTableUtil.createOptimizingPlanner(
@@ -363,7 +354,7 @@ public class OptimizingQueue extends PersistentBase {
         return null;
       }
     } catch (Throwable throwable) {
-      tableRuntime.planFailed();
+      tableRuntime.planFailed(ExceptionUtil.getErrorMessage(throwable, 4000), throwable);
       LOG.error("Planning table {} failed", tableRuntime.getTableIdentifier(), throwable);
       throw throwable;
     }
@@ -854,16 +845,15 @@ public class OptimizingQueue extends PersistentBase {
               doAs(
                   TableProcessMapper.class,
                   mapper ->
-                      mapper.insertProcess(
+                      mapper.updateProcess(
                           tableRuntime.getTableIdentifier().getId(),
                           processId,
                           "",
                           status,
-                          optimizingType.name().toUpperCase(),
                           tableRuntime.getOptimizingStatus().name().toLowerCase(),
-                          "AMORO",
                           0,
-                          planTime,
+                          0L,
+                          "",
                           new HashMap<>(),
                           getSummary().summaryAsMap(false))),
           () ->
@@ -910,7 +900,7 @@ public class OptimizingQueue extends PersistentBase {
           () -> tableRuntime.completeProcess(success),
           () -> clearProcess(this));
       if (!success && failedReason != null) {
-        appendFailReasonToDriverLogs(failedReason, processId);
+        tableRuntime.appendFailReasonToDriverLogs(failedReason, processId);
       }
     }
 
@@ -962,33 +952,6 @@ public class OptimizingQueue extends PersistentBase {
         taskRuntime.getCompletedFuture().whenCompleted(() -> acceptResult(taskRuntime));
         taskMap.put(taskRuntime.getTaskId(), taskRuntime);
         taskQueue.offer(taskRuntime);
-      }
-    }
-
-    private void appendFailReasonToDriverLogs(String failedReason, long processId) {
-      String envLogDir = System.getenv("LOG_DIR");
-      String logBaseDir =
-          (envLogDir != null && !envLogDir.isEmpty()) ? envLogDir : "/mnt/amoro-logs/compaction";
-      Path driverLogPath = Paths.get(logBaseDir, String.valueOf(processId), "driver.log");
-      String time =
-          DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS")
-              .withZone(ZoneOffset.UTC)
-              .format(Instant.now());
-      try {
-        Files.createDirectories(driverLogPath.getParent());
-        Map<String, String> logEntry = new LinkedHashMap<>();
-        logEntry.put("level", "ERROR");
-        logEntry.put("time", time);
-        logEntry.put("processId", String.valueOf(processId));
-        logEntry.put("taskId", "");
-        logEntry.put("logger", "");
-        logEntry.put("message", failedReason);
-        logEntry.put("stackTrace", "");
-        String logLine = new ObjectMapper().writeValueAsString(logEntry) + System.lineSeparator();
-        Files.writeString(
-            driverLogPath, logLine, StandardOpenOption.CREATE, StandardOpenOption.APPEND);
-      } catch (Exception e) {
-        LOG.warn("Failed to append fail reason to driver log for process {}", processId, e);
       }
     }
   }
--- a/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
@@ -909,7 +909,7 @@ public class OptimizingQueue extends PersistentBase {
           () -> tableRuntime.completeProcess(success),
           () -> clearProcess(this));
       if (!success && failedReason != null) {
-        tableRuntime.appendFailReasonToDriverLogs(failedReason, processId);
+        tableRuntime.appendDriverLogEntry(processId, "ERROR", failedReason, null);
       }
     }
 
--- a/amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
@@ -111,7 +111,6 @@ public class TableRuntimeRefreshExecutor extends PeriodicTableScheduler {
       defaultTableRuntime.refresh(table);
 
       MixedTable mixedTable = (MixedTable) table.originalTable();
-      // first, evaluate cron then update healthscore
       evaluateCronTriggers(defaultTableRuntime, mixedTable);
     } catch (Throwable throwable) {
       logger.error("Refreshing table {} failed.", tableRuntime.getTableIdentifier(), throwable);
--- a/amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
@@ -113,7 +113,6 @@ public class TableRuntimeRefreshExecutor extends PeriodicTableScheduler {
       MixedTable mixedTable = (MixedTable) table.originalTable();
       // first, evaluate cron then update healthscore
       evaluateCronTriggers(defaultTableRuntime, mixedTable);
-      updateHealthScore(defaultTableRuntime, mixedTable);
     } catch (Throwable throwable) {
       logger.error("Refreshing table {} failed.", tableRuntime.getTableIdentifier(), throwable);
     }
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
@@ -64,15 +64,15 @@ import org.apache.iceberg.Snapshot;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.nio.file.Files;
+import java.util.HashMap;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.nio.file.StandardOpenOption;
-import java.time.Instant;
-import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
-import java.util.HashMap;
+import java.time.ZoneOffset;
+import java.time.Instant;
 import java.util.LinkedHashMap;
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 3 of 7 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -854,16 +845,15 @@ private void beginAndPersistProcess() {
               doAs(
                   TableProcessMapper.class,
                   mapper ->
-                      mapper.insertProcess(
+                      mapper.updateProcess(
                           tableRuntime.getTableIdentifier().getId(),
                           processId,
                           "",
                           status,
-                          optimizingType.name().toUpperCase(),
```
badalprasadsingh, line 862: Switching from `insert` to `update` is correct here.

But we skipped `process_type`, this makes AMS have the wrong / assumed compaction type. Please add it.

```java
          () ->
              doAs(
                  TableProcessMapper.class,
                  mapper ->
                      mapper.updateProcessType(
                          tableRuntime.getTableIdentifier().getId(),
                          processId,
                          optimizingType.name().toUpperCase())),
```

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -362,18 +378,58 @@ public DefaultTableRuntime refresh(AmoroTable<?> table) {
...
+
+    return processId;
   }
 
-  public void planFailed() {
-    OptimizingStatus originalStatus = getOptimizingStatus();
-    store().begin().updateStatusCode(code -> OptimizingStatus.IDLE.getCode()).commit();
+  public void planFailed(String reason) {
+    planFailed(reason, null);
+  }
```
badalprasadsingh: This function simply calls another function, just below it.

Drop it and use the single function.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -466,21 +522,99 @@ public void completeEmptyProcess() {
...
+        .updateState(PENDING_INPUT_KEY, any -> new AbstractOptimizingEvaluator.PendingInput())
+        .updateState(PROCESS_ID_KEY, any -> 0L)
+        .commit();
+    this.pendingCronType = null;
+  }
+
+  private void finalizePlanningProcess(
+      long processId, ProcessStatus status, String reason, Throwable throwable) {
+    long now = System.currentTimeMillis();
+    OptimizingType type = pendingCronType != null ? pendingCronType : OptimizingType.MINOR;
```
badalprasadsingh: Here: if this function throws the error and functions like `completeEmptyProcess` / `planFailed` which calls it before updating the status to IDLE, are forced to exit before updating the state.

Because, of it, the process will be stuck.

Add `doAS` under try / catch block logging out the issue. You can decide where it should be warning or error.

amoro-ams/src/main/java/org/apache/amoro/server/utils/IcebergTableUtil.java
```diff
@@ -284,7 +286,10 @@ public static AbstractOptimizingPlanner createOptimizingPlanner(
                             table, entry.getKey(), entry.getValue()))
                 .reduce(Expressions::or)
                 .orElse(Expressions.alwaysTrue());
-    long processId = snowflakeIdGenerator.generateId();
+    long processId =
+        tableRuntime.getProcessId() != 0
+            ? tableRuntime.getProcessId()
+            : snowflakeIdGenerator.generateId();
```
badalprasadsingh: Let's not read the `processId` from the runtime. Problem: runtime behaves as a cache. If any process type is finished and committed, we will reuse the older id here.

Do this: get the process id earlier in `planInternal` function itself : `tableRuntime.getProcessId()` and pass it to this `createOptimizingPlanner` function by adding the param in the function for it.

badalprasadsingh: Thus simply drop this line.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -466,21 +522,99 @@ public void completeEmptyProcess() {
...
+        String.format(
+            "cron fired for %s but planner evaluated and found no data to process",
+            cronType != null ? cronType : "UNKNOWN");
+
+    long processId = getProcessId();
+    if (processId != 0) {
+      finalizePlanningProcess(processId, ProcessStatus.SKIPPED, reason, null);
+    } else if (cronType != null) {
+      recordSkippedOptimization(cronType, reason);
+    }
```
badalprasadsingh: well, this path of getting a non-zero `processId` is unreachable, as non of its callers ever pass a non-zero value for processId.

drop the second else if case else if case.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -79,6 +90,11 @@ public class DefaultTableRuntime extends AbstractTableRuntime
 
   private static final SnowflakeIdGenerator ID_GENERATOR = new SnowflakeIdGenerator();
 
+  private static final String DRIVER_LOG_LOGGER = "OptimizingQueue";
```
badalprasadsingh: we have the field of `JSON_LOG_PATTERN` in `conf/optimize/log4j2-routing.xml` (I only added it).

This must stay identical to it as the optimizer executor appends to the same `driver.log`.

Drop it simply and simply let its caller remain empty `logEntry.put("logger", "");`

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -362,18 +378,58 @@ public DefaultTableRuntime refresh(AmoroTable<?> table) {
     return this;
   }
 
-  public void beginPlanning() {
-    OptimizingStatus originalStatus = getOptimizingStatus();
-    store().begin().updateStatusCode(code -> OptimizingStatus.PLANNING.getCode()).commit();
+  public long beginPlanning() {
```
badalprasadsingh: No need to return processId / anything from here. Change it to `void`.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -362,18 +378,58 @@ public DefaultTableRuntime refresh(AmoroTable<?> table) {
...
+    long processId = ID_GENERATOR.generateId();
+    long now = System.currentTimeMillis();
+    OptimizingType type = pendingCronType != null ? pendingCronType : OptimizingType.MINOR;
+    Map<String, String> summary = new HashMap<>();
+    summary.put("optimizingType", type.name());
+    summary.put("phase", "PLANNING");
```
badalprasadsingh: IMO (not sure), this is redundant as it gets override later, we can drop this line: `summary.put("phase", "PLANNING");`

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -362,18 +378,58 @@ public DefaultTableRuntime refresh(AmoroTable<?> table) {
     return this;
   }
 
-  public void beginPlanning() {
-    OptimizingStatus originalStatus = getOptimizingStatus();
-    store().begin().updateStatusCode(code -> OptimizingStatus.PLANNING.getCode()).commit();
+  public long beginPlanning() {
+    long processId = ID_GENERATOR.generateId();
+    long now = System.currentTimeMillis();
+    OptimizingType type = pendingCronType != null ? pendingCronType : OptimizingType.MINOR;
```
badalprasadsingh: Its wrong to add a fallback that if the cron type is `null`, then make it `MINOR`.

Create one utility function like this and reuse everywhere.
```java
  private String triggeredTypeName() {
    OptimizingType cronType = this.pendingCronType;
    return cronType == null ? "UNKNOWN" : cronType.name();
  }
```

badalprasadsingh: And reuse the process type returned from it wherever required.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -466,21 +522,99 @@ public void completeEmptyProcess() {
...
+        .begin()
+        .updateStatusCode(code -> OptimizingStatus.IDLE.getCode())
+        .updateState(PENDING_INPUT_KEY, any -> new AbstractOptimizingEvaluator.PendingInput())
+        .updateState(PROCESS_ID_KEY, any -> 0L)
+        .commit();
+    this.pendingCronType = null;
+  }
+
+  private void finalizePlanningProcess(
+      long processId, ProcessStatus status, String reason, Throwable throwable) {
```
badalprasadsingh: nit and non-blocking: remove the processId param from the function and get the value inside the function itself.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -466,21 +522,99 @@ public void completeEmptyProcess() {
...
+    long now = System.currentTimeMillis();
```
badalprasadsingh, line 544: Lets add this check here, the only time processId can be zero is when AMS restarted and lost the context.

```java
    if (getOptimizingStatus() != OptimizingStatus.PLANNING || processId == 0) {
      return;
    }
```

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -466,21 +522,99 @@ public void completeEmptyProcess() {
...
+    OptimizingType type = pendingCronType != null ? pendingCronType : OptimizingType.MINOR;
+    Map<String, String> summary = new HashMap<>();
+    summary.put("optimizingType", type.name());
```
badalprasadsingh: call the utility function I told you about, and use it here : `triggeredTypeName`

badalprasadsingh (changes requested):
First pass. The fix looks good. Two major misses, few inlines and regarding code quality.

Also, have we compiled it? along with running the tests, as the tests might break (please just check that once.)

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
ashisharma-dz: insertProcess() apparently expects a map<string, string> as a summary, hence this

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
ashisharma-dz: would need confirmation on this though:
this version of planFailed does have a caller right now,
planFailed is being called without any throwable right now in initTableRuntime() OptimizingQueue.java
with the purpose of resetting a table stranded in planning.

should we directy update it to IDLE? or should be marked as failed? right now planFailed requires a throwable as parameter for appending the log

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -466,21 +513,107 @@ public void completeEmptyProcess() {
...
+      return;
+    }
+
+    String reason =
+        String.format(
+            "cron fired for %s but planner evaluated and found no data to process",
+            triggeredTypeName());
+
+    finalizePlanningProcess(ProcessStatus.SKIPPED, reason, null);
+    recordSkippedOptimization(this.pendingCronType, reason);
```
badalprasadsingh: We should not call `recordSkippedOptimization` in this function again here. This will store another entry in the DB, we are already storing once.

Drop the line calling `recordSkippedOptimization`.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -362,18 +376,51 @@ public DefaultTableRuntime refresh(AmoroTable<?> table) {
...
+    summary.put("optimizingType", triggeredTypeName());
+    doAs(
+        TableProcessMapper.class,
+        mapper ->
+            mapper.insertProcess(
+                getTableIdentifier().getId(),
+                processId,
+                "",
+                ProcessStatus.RUNNING,
+                triggeredTypeName().toUpperCase(),
```
badalprasadsingh: nit(non-blocking): call `triggeredTypeName` once, and reuse its value at both the call sites.

badalprasadsingh (changes requested):
Minor Comments. Also, please sign the `CLA` in case not signed (check the comment right above).

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -348,22 +340,22 @@ private TableOptimizingProcess planInternal(DefaultTableRuntime tableRuntime) {
...
+              tableRuntime.getProcessId());
       if (planner.isNecessary()) {
         return new TableOptimizingProcess(planner, tableRuntime);
       } else {
         tableRuntime.completeEmptyProcess();
         return null;
       }
     } catch (Throwable throwable) {
-      tableRuntime.planFailed();
+      tableRuntime.planFailed(ExceptionUtil.getErrorMessage(throwable, 4000), throwable);
```
badalprasadsingh: We are limiting the stack trace to `4000`, lets not pass a value then the complete stack trace will be sent.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -466,21 +514,106 @@ public void completeEmptyProcess() {
...
+    }
+
+    if (status == ProcessStatus.FAILED) {
+      appendDriverLogEntry(processId, "ERROR", describeThrowable(throwable, reason), throwable);
+    }
+  }
+
+  public void appendFailReasonToDriverLogs(String failedReason, long processId) {
+    appendDriverLogEntry(processId, "ERROR", failedReason, null);
+  }
```
badalprasadsingh: nit: unnecessary function, please reuse  `appendDriverLogEntry`

badalprasadsingh (commented):
Code wise approved. Not tested. Minor non-blocking nit.

hash-data:
AI review 
```

P1 — TableRuntimeRefreshExecutor.java:114: health score / table summary update deleted, unrelated to PR. Commit 28a971879 ("header and minor commit") dropped updateHealthScore(defaultTableRuntime, mixedTable). updateHealthScore (line 171) is now a dead private method — only production caller of tableRuntime.setTableSummary(...) (line 180). Result: TableSummary never refreshes after DefaultTableService:502 seeds new TableSummary(), so TableController:168, OverviewManager:215 and gauge TableSummaryMetrics:204 report a frozen health score forever. Restore the call, or split the removal into its own PR with rationale and delete the dead method + maxPendingPartitions.
Self-challenge: could be a deliberate perf fix — createOptimizingEvaluatorresh cycle. But nothing else sets the summary, no commit message or testsays so, and the dead method was left in place. Stands.                                                                                                              
P2 — DefaultTableRuntime.java:537-570: zombie RUNNING rows, no reconciliation. Two windows: (a) finalizePlanningProcess catches Exception around updateProcess, logs the caller still flips status to IDLE — the row stays RUNNING forever; (b)ertProcess in one doAs, then commits status/PROCESS_ID_KEY in a separatestore().commit() — crash between leaves a RUNNING row on a table that is never PLANNING. Startup recovery (OptimizingQueue.initTableRuntime:134) only fires on statusPLANNING, so neither is cleaned; OptimizingExpiringExecutor only removes tn startup, mark table_process rows still RUNNING whose table is notprocessing as FAILED.                                                                                                                                                Self-challenge: swallowing the DB error is arguably right (a DB blip shoulLANNING), and expiry bounds the leak. That's why this is P2, not P1 — butthe PR's whole point is accurate process status, so a reconciliation pass belongs in it.                                                                             
P2 — DefaultTableRuntime.java:378-381: process_type = 'UNKNOWN' after AMS restart. pendingCronType is in-memory (volatile, line 124) while OptimizingStatus.PENDING ipersisted. After restart, SchedulingPolicy picks the PENDING table, triggeNOWN", and the row is inserted with that. If planning succeedsupdateProcessType corrects it in the same transaction; if it is SKIPPED or FAILED, UNKNOWN is what the dashboard shows (MixedAndIcebergTableDescriptor:1027 setOptimiPersist the cron type in the runtime state alongside PENDING, or derive thtor.
Checked and cleared: TableOptimizingProcess(tableRuntime, meta, state):502 does OptimizingType.valueOf(processMeta.getProcessType()), which would throw on UNKNOWN — unreachable, because updateProcessType and beginProcess share one doAsTranso no processing-state row can carry UNKNOWN.

P2 — OptimizingQueue.java:843-857: insertProcess → updateProcess makes rowecondition. beginAndPersistProcess now assumes beginPlanning inserted therow. updateProcess returns void; a 0-row UPDATE is silent, and the transaction proceeds to insert optimizing_process_state + task_runtime for a process with no table_process
row. That state is unrecoverable: initTableRuntime:148 would get meta == n.getProcessType() at startup. Have the mapper return int and fail loudlywhen it is 0.

P2 — test coverage. Persistence lifecycle changed (insert point, status transitions, id ownership) with zero new tests; TestOptimizingPlanner only gained a signature argument.
Worth one test over beginPlanning → planFailed / completeEmptyProcess assenish_time, and fail_message.

P3 — DefaultTableRuntime.java:583: new ObjectMapper() per log line. Hoist ame as DRIVER_LOG_TIME_FORMATTER.

No P0. finish_time = 0L is fine — Long2TsConverter:31 maps 0 to SQL NULL. , so the 4000-char stack trace fits. Row growth from per-tick SKIPPED rowsis bounded — OptimizingExpiringExecutor expires by time, not count.
```

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -466,21 +514,102 @@ public void completeEmptyProcess() {
...
+    store()
+        .begin()
+        .updateStatusCode(code -> OptimizingStatus.IDLE.getCode())
+        .updateState(PENDING_INPUT_KEY, any -> new AbstractOptimizingEvaluator.PendingInput())
+        .updateState(PROCESS_ID_KEY, any -> 0L)
+        .commit();
+    this.pendingCronType = null;
+  }
+
+  private void finalizePlanningProcess(ProcessStatus status, String reason, Throwable throwable) {
```
hash-data, line 538: let us add comments on functions

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -466,21 +514,102 @@ public void completeEmptyProcess() {
...
+            "cron fired for %s but planner evaluated and found no data to process",
+            triggeredTypeName());
+
+    finalizePlanningProcess(ProcessStatus.SKIPPED, reason, null);
+
+    store()
+        .begin()
+        .updateStatusCode(code -> OptimizingStatus.IDLE.getCode())
+        .updateState(PENDING_INPUT_KEY, any -> new AbstractOptimizingEvaluator.PendingInput())
+        .updateState(PROCESS_ID_KEY, any -> 0L)
```
hash-data, line 532: what we are updating here?

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -362,18 +376,52 @@ public DefaultTableRuntime refresh(AmoroTable<?> table) {
...
+        TableProcessMapper.class,
+        mapper ->
+            mapper.insertProcess(
+                getTableIdentifier().getId(),
+                processId,
+                "",
+                ProcessStatus.RUNNING,
+                optimizingTypeName.toUpperCase(),
+                OptimizingStatus.PLANNING.name().toLowerCase(),
+                "AMORO",
```
hash-data: do we need to use AMORO ?

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -362,18 +376,52 @@ public DefaultTableRuntime refresh(AmoroTable<?> table) {
...
+                0,
```
hash-data, line 400: also how retry number is decided?

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
```diff
@@ -111,9 +111,7 @@ public void execute(TableRuntime tableRuntime) {
       defaultTableRuntime.refresh(table);
 
       MixedTable mixedTable = (MixedTable) table.originalTable();
-      // first, evaluate cron then update healthscore
       evaluateCronTriggers(defaultTableRuntime, mixedTable);
-      updateHealthScore(defaultTableRuntime, mixedTable);
```
hash-data, line 116: if we remove this we will not get health score on ui?

badalprasadsingh:
image

Its working fine, but we need to debug the case we should remove any case where we get "UNKNOWN" as opt type.

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
ashisharma-dz: yes. we won't. there is no other evaluation of healthscore right now.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
ashisharma-dz: "AMORO"  (the process execution engine), is required by the insertProcess() API

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
ashisharma-dz: the TableRuntimeStore for each table stores the current process_id, we had set it to the process_id in when planning began, here we are overwriting it with 0.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
ashisharma-dz: I think this is not the retry number, but more of a retry count, initialised to 0 on each insertiuon of process. Technically, this should get incremented everytime the same process goes into retry. This is part of the insertProcess() API framework, but not really an implementation we use.

hash-data (approved):

badalprasadsingh (approved):

PR #32, opened 2026-09-07
feat(ams): database as source of truth for iceberg table configurations

### LINKED ISSUE #28: [Feature]: Make DB as the Single Source of Truth for Iceberg Table Metadata
User level configurations such as `enable` / `disable`, `cron` timings & `target-file-size` play a important part, as they are continuously **fine-tuned** and set.

Now, every time the user runs **Clear Destination** via **OLake-Go**, we lose out on this important configuration values, as they are stored in `metadata.json` file of the iceberg table.

So the **proposed solution** is: to store these values in the DB, so it can be picked up again, even if the iceberg table is dropped. _Note:_ the new table must be recreated with the same table-name and database-name.

**Bonus:** This will also reduce the time taken to fetch all iceberg tables as they now would be stored in db.

badalprasadsingh:
Fixes #28, from now on, `db` will act as the single source of truth, no cache or reading from `metadata.json`


badalprasadsingh:
This PR is on the telemetry branch, so that must go before this.

amoro-ams/src/main/resources/postgres/migration/V2__table_configurations.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE IF NOT EXISTS table_configurations
```
ashisharma-dz: We might have to add this to migrations in `SchemaMigrator.java`, do check once:  (L86)
 ` private static final String[] MIGRATIONS = {"V1__platform_property.sql"};`

amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurationsService.java
```diff
@@ -0,0 +1,213 @@
...
+
+  private void persist(TableOptimizingConfigurationsMeta meta) {
+    doAs(
+        TableConfigurationsMapper.class,
+        mapper -> {
+          if (mapper.updateSettings(meta) == 0) {
+            mapper.insertSettings(meta);
+          }
+        });
+  }
```
ashisharma-dz: do check this again wrt: 
persist() updates every column. storeHealthScore() and update() each load a row, change their own fields, and write the whole row back. A health tick that loaded the row before a config save will write the old crons back.

amoro-ams/src/main/resources/postgres/migration/V2__table_configurations.sql
```diff
@@ -0,0 +1,17 @@
...
+(
+    catalog_name             VARCHAR(64) NOT NULL,
+    db_name                  VARCHAR(128) NOT NULL,
+    table_name               VARCHAR(256) NOT NULL,
+    self_optimizing_enabled  BOOLEAN NOT NULL DEFAULT FALSE,
+    minor_trigger_cron       VARCHAR(128) DEFAULT NULL,
+    major_trigger_cron       VARCHAR(128) DEFAULT NULL,
+    full_trigger_cron        VARCHAR(128) DEFAULT NULL,
+    target_size              BIGINT NOT NULL DEFAULT 536870912,
```
ashisharma-dz: `self_optimizing_enabled` and `target_size` default values set in the database (here and through `adopt()`) are inconsistent with the original default values in `TableProperties.java`: TRUE and 128MB, can we make both defaults same?

amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurationsService.java
```diff
@@ -0,0 +1,213 @@
...
+  // either returns the existing table configurations for an iceberg table from db, or,
+  // stores the entry in db for the new iceberg table, and returns
+  public TableOptimizingConfigurationsMeta getOrCreate(ServerTableIdentifier identifier) {
+    TableOptimizingConfigurationsMeta existing = select(identifier);
+    if (existing != null) {
+      return existing;
+    }
+    TableOptimizingConfigurationsMeta meta =
+        new TableOptimizingConfigurationsMeta(
+            identifier.getCatalog(), identifier.getDatabase(), identifier.getTableName());
```
ashisharma-dz: use a helper function for this instantiation:


```
private static TableOptimizingConfigurationsMeta newMeta(ServerTableIdentifier id) {
  return new TableOptimizingConfigurationsMeta(id.getCatalog(), id.getDatabase(), id.getTableName());
}
```

use this wherever this repeats (have marked such parts with comments)

amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurationsService.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  public void update(
+      Collection<ServerTableIdentifier> identifiers, TableOptimizingConfigurationsMeta values) {
+    for (ServerTableIdentifier identifier : identifiers) {
+      TableOptimizingConfigurationsMeta meta = select(identifier);
+      if (meta == null) {
+        meta =
+            new TableOptimizingConfigurationsMeta(
+                identifier.getCatalog(), identifier.getDatabase(), identifier.getTableName());
```
ashisharma-dz: use the helper function here

amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurationsService.java
```diff
@@ -0,0 +1,213 @@
...
+    LOG.info("Adopted optimizing properties for iceberg tables");
+  }
+
+  private void adoptTableProperties(ServerTableIdentifier identifier, AmoroTable<?> table) {
+    UnkeyedTable unkeyed = (UnkeyedTable) table.originalTable();
+    Map<String, String> stored = ((HasTableOperations) unkeyed).operations().current().properties();
+
+    TableOptimizingConfigurationsMeta meta =
+        new TableOptimizingConfigurationsMeta(
+            identifier.getCatalog(), identifier.getDatabase(), identifier.getTableName());
```
ashisharma-dz: helper function to be used here

amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurationsService.java
```diff
@@ -0,0 +1,213 @@
...
+            TableProperties.SELF_OPTIMIZING_TARGET_SIZE,
+            TableOptimizingConfigurationsMeta.DEFAULT_TARGET_SIZE));
+  }
+
+  public void storeHealthScore(ServerTableIdentifier identifier, int healthScore, long snapshotId) {
+    TableOptimizingConfigurationsMeta meta = select(identifier);
+    if (meta == null) {
+      meta =
+          new TableOptimizingConfigurationsMeta(
+              identifier.getCatalog(), identifier.getDatabase(), identifier.getTableName());
```
ashisharma-dz: helper function here

amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurationsService.java
ashisharma-dz: the same "select, else make new" logic is here and in `storeHealthScore()`, and in `getOrCreate()`
common it out in a function like this:

`private TableOptimizingConfigurationsMeta selectOrNew(ServerTableIdentifier id) {
  TableOptimizingConfigurationsMeta meta = select(id);
  return meta != null ? meta : newMeta(id);
}
`

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/TableConfigurationDetails.java
```diff
@@ -0,0 +1,108 @@
...
+    String db = ctx.pathParam("db");
+    check(catalog, db);
+
+    OptimizingConfigurations info = ctx.bodyAsClass(OptimizingConfigurations.class);
+
+    List<ServerTableIdentifier> identifiers = Lists.newArrayList();
+    for (String tableName : info.getTables()) {
+      ServerTableIdentifier identifier =
+          tableManager.getServerTableIdentifier(
+              TableIdentifier.of(catalog, db, tableName).buildTableIdentifier());
```
ashisharma-dz: extract `private ServerTableIdentifier resolve(String catalog, String db, String table)` and call it from both `updateConfigurations()` and `getTableConfig()`

amoro-ams/src/main/java/org/apache/amoro/server/persistence/mapper/TableConfigurationsMapper.java
```diff
@@ -0,0 +1,116 @@
...
+  int updateSettings(TableOptimizingConfigurationsMeta meta);
+
+  /* ---------- insert ---------- */
+
+  @Insert(
+      "INSERT INTO "
+          + TABLE_NAME
+          + " (catalog_name, db_name, table_name, self_optimizing_enabled, "
+          + "  minor_trigger_cron, major_trigger_cron, full_trigger_cron, target_size, "
+          + "  olake_created, health_score, health_score_snapshot_id) "
```
ashisharma-dz: same as the SELECT_COLS list above. We can use that here.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/TableConfigurationDetails.java
```diff
@@ -0,0 +1,108 @@
...
+    for (ServerTableIdentifier identifier : tableManager.listIcebergTables(catalog, db)) {
+      result.add(configurations.getOrCreate(identifier));
+    }
+    ctx.json(OkResponse.of(result));
+  }
+
+  public void getTableConfig(Context ctx) {
+    String catalog = ctx.pathParam("catalog");
+    String db = ctx.pathParam("db");
+    String table = ctx.pathParam("table");
```
ashisharma-dz: we can make "catalog", "table", and "db" as constant literals since these keys are used here multiple times, as in:
`private static final String CATALOG_KEY = "catalog"`

ashisharma-dz (changes requested):
First Review
most comments remain on redundancy

- one important thing to check: if some tests might break in runtime due to the pending derby and mysql tables. 
- we don't use derby anymore in our current merge, check on the TO DO once
- V2_table_configurations.sql needs to be added to the migrations list
- add the license header to the files flagged in check fail

badalprasadsingh:
Right now, `chore/migration_and_telemetry_changes` is way behind `origin/staging`. Merged staging to it, brings about a lot of unneccessary changes because of recent vulnebarility fixes.

Change the origin branch to staging for now for ease of review.

amoro-ams/src/main/resources/postgres/migration/V2__table_configurations.sql
badalprasadsingh: Valid. Reverted to staging branch, so that is gone. Will do it after telemetry is merged.

amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurationsService.java
badalprasadsingh: Nice. Fixed.

amoro-ams/src/main/resources/postgres/migration/V2__table_configurations.sql
badalprasadsingh: Intentional. Fusion's default values are different from Amoro.

amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurationsService.java
badalprasadsingh: Nice. Done.

amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurationsService.java
badalprasadsingh: Done.

amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurationsService.java
badalprasadsingh: Done.

amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurationsService.java
badalprasadsingh: Done.

amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurationsService.java
badalprasadsingh: Not needed anymore.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/TableConfigurationDetails.java
badalprasadsingh: Done.

amoro-ams/src/main/java/org/apache/amoro/server/persistence/mapper/TableConfigurationsMapper.java
badalprasadsingh: Done.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/TableConfigurationDetails.java
badalprasadsingh: Done.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/TableConfigurationDetails.java
```diff
@@ -60,24 +79,89 @@ public void getIcebergTables(Context ctx) {
     String db = ctx.pathParam(DB_KEY);
     check(catalog, db);
 
-    List<TableOptimizingConfigurationsMeta> result = Lists.newArrayList();
-    for (ServerTableIdentifier identifier : tableManager.listIcebergTables(catalog, db)) {
-      result.add(configurations.getOrCreate(identifier));
+    // the tables the catalog holds right now, with the catalog's database and table filters leaving
+    // out the same tables AMS leaves out when it syncs the catalog
+    ExternalCatalog serverCatalog = (ExternalCatalog) catalogManager.getServerCatalog(catalog);
```
ashisharma-dz: the unconditional (ExternalCatalog) cast throws ClassCastException (HTTP 500) for an internal catalog
maybe use an instanceof check

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/TableConfigurationDetails.java
```diff
@@ -60,24 +79,89 @@ public void getIcebergTables(Context ctx) {
...
 
   public void getTableConfig(Context ctx) {
-    String catalog = ctx.pathParam(CATALOG_KEY);
-    String db = ctx.pathParam(DB_KEY);
-    String table = ctx.pathParam(TABLE_KEY);
-    check(catalog, db);
-    Preconditions.checkArgument(StringUtils.isNotBlank(table), "table can not be empty");
-
-    ServerTableIdentifier identifier = resolve(catalog, db, table);
-    Preconditions.checkArgument(identifier != null, "Unknown table %s.%s.%s", catalog, db, table);
```
ashisharma-dz: check this flagged by ai once:
the old "Unknown table" precondition is gone.
   - updateConfigurations with a misspelled name makes persist() insert a row for a table that does not exist.
   - getTableConfig for an unknown table fails in loadTable and returns 500, not 400.

ashisharma-dz (changes requested):
mostly approved. just check these 2 comments

badalprasadsingh:
- It adds the health-score logic removed in plan failed PR. It now works as an independent process, if it fails it breaks nothing
- 4 new APIs introduced in Fusion
- slight change in api contract in olake-ui
- OLake-UI will now use Fusion APIs using piggybacking, its older APIs will be removed
- these new apis are faster as they check db instead of iceberg files (expensive s3 requests)
- On startup configurations from older iceberg tables are copied and stored in db (backward compatibility)
- migration logic added and dependent on telemetry PR
- table configurations are not lost even if the iceberg table is dropped from the catalog, only lost when the catalog is dropped from fusion

PR #35, opened 2026-09-09
fix: bump netty 4.1.138

ashisharma-dz:

Critical Vulnerability 

Close #221.

## Brief change log

-bumped netty in pom.xml
-regenerated dev/deps/dependencies-hadoop-3-spark-3.5 and dev/deps/dependencies-hadoop-2-spark-3.3


badalprasadsingh (dismissed):

hash-data:
AI review: 
```

P0

1. 4.1.137.Final is itself vulnerable. Target 4.1.138.Final.
CVE-2026-89044 — HttpObjectDecoder fails to validate the final transfer coding, so Transfer-Encoding: chunked, xchunked or a split TE header decodes as chunked anyway. Request smuggling, CVSS 6.5. Affected range is 4.1.133 → 4.1.137; fixed in 4.1.138. It was introduced by the CVE-2026-42581 fix in 4.1.133.

Current 4.1.129 is not in that range. So this PR removes old CVEs and adds a new one. For a bump whose entire purpose is scanner hygiene, 4.1.137 gets flagged the day it lands. One-character fix at pom.xml:145.

Self-challenge: real exploitability here is near zero (see P1-3 — nothing first-party decodes HTTP with netty). But the PR's goal is the scanner, and it misses.

2. Spark 3.5.8 is silently moved 41 patch versions off its tested netty.
spark-parent_2.13:3.5.8 declares <netty.version>4.1.96.Final</netty.version>. netty-all (direct dep at amoro-ams/pom.xml:249-252) drags all ~35 netty modules at ${netty.version} and wins nearest-first, so Spark's shuffle and RPC transport run on 4.1.137. That is the largest blast radius in this PR and it appears nowhere in the commit message.

Not hypothetical: netty 4.1.130 was released specifically to fix a regression caused by 4.1.129, and 4.1.125 shipped a severe throughput regression. Needs a shuffle-heavy soak test before merge, not just a green unit build.
```

PR #36, opened 2026-09-09
fix: bump trino avro to 1.11.5

ashisharma-dz:

Critical vulnerability

Close  Alert #177.

## Brief change log

-amoro-mixed-trino/pom.xml, avro version bumped to 1.11.4

### DIFF
--- a/amoro-format-mixed/amoro-mixed-trino/pom.xml
+++ b/amoro-format-mixed/amoro-mixed-trino/pom.xml
@@ -15,6 +15,8 @@
   ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
+  ~
+  ~ Modified by Datazip Inc. in 2026
   -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
--- a/amoro-format-mixed/amoro-mixed-trino/pom.xml
+++ b/amoro-format-mixed/amoro-mixed-trino/pom.xml
@@ -41,7 +41,7 @@
         <inject.version>1</inject.version>
         <servlet-api.version>4.0.1</servlet-api.version>
         <commons-collections.version>3.2.2</commons-collections.version>
-        <avro.version>1.10.1</avro.version>
+        <avro.version>1.11.4</avro.version>
         <jersey.version>1.19.4</jersey.version>
         <guava-retrying.version>2.0.0</guava-retrying.version>
         <commons-lang.version>2.6</commons-lang.version>


badalprasadsingh:
Approved!

badalprasadsingh (dismissed):

hash-data:
AI Review 
```

P1

1. 1.11.4 is still vulnerable — use 1.11.5.
CVE-2025-33042 (code injection when generating specific records from untrusted schemas, CVSS 7.3, published 2026-08-24) affects all versions through 1.11.4, plus 1.12.0. Fixed in 1.11.5 / 1.12.1.

This is the identical pattern to the netty branch: a security bump that lands inside the affected range of a newer CVE. Both PRs will be re-flagged by whatever scanner prompted them.

1.11.5 is the right target, not 1.12.x — and the reason is Jackson:

┌─────────────┬──────────────────────────────────┐
│ avro-parent │          jackson floor           │
├─────────────┼──────────────────────────────────┤
│ 1.11.4      │ 2.14.3                           │
├─────────────┼──────────────────────────────────┤
│ 1.11.5      │ 2.14.3 ← same, zero new pressure │
├─────────────┼──────────────────────────────────┤
│ 1.12.0      │ 2.17.2                           │
├─────────────┼──────────────────────────────────┤
│ 1.12.1      │ 2.20.0                           │
└─────────────┴──────────────────────────────────┘

The module pins <jackson.version>2.13.5</jackson.version> (pom.xml:41). Moving to 1.12.x would open a seven-minor Jackson gap. 1.11.5 is a one-character change carrying no new risk.

Real exploitability here is ≈0 — the CVE is about SpecificCompiler/specific-record generation, and this module only uses GenericDatumReader, with no codegen. But the bump's purpose is scanner hygiene, so it should land on the fixed version.

2. Avro now diverges across the repo for the same Iceberg.
avro.version is defined in exactly one place repo-wide — this module. Everywhere else avro appears only as an <exclusion>. The main dist resolves avro transitively to 1.12.0 (via iceberg-core:1.7.2), confirmed by the netty branch's manifests. After this PR, Iceberg 1.7.2 runs on avro 1.12.0 in the dist and 1.11.4 in the Trino plugin.

That's defensible — they're separately-classloaded artifacts with different Jackson constraints — but it should be a deliberate, commented decision rather than an accident, or the next person will "helpfully" align them and break the Jackson floor.

P2

3. Jackson relationship inverted, and already split. Resolved tree for the module:
+- io.trino:trino-plugin-toolkit:jar:406:compile
|  \- com.fasterxml.jackson.core:jackson-core:jar:2.14.2:compile
+- com.fasterxml.jackson.core:jackson-databind:jar:2.13.5:compile
+- org.apache.avro:avro:jar:1.11.4:compile
\- com.fasterxml.jackson.core:jackson-annotations:jar:2.13.5:provided
jackson-core 2.14.2 against jackson-databind 2.13.5 is a pre-existing split (core-newer is the tolerated direction). What this PR changes is the avro side: avro 1.10.1 needed databind 2.11.3 and got 2.13.5 — comfortably above floor. Avro 1.11.4 wants 2.14.3 and gets 2.13.5 — now below.

Low probability (Avro uses stable JsonNode/ObjectMapper APIs; StreamReadConstraints is 2.15+ so it can't be implicated), and the tests pass — but the failure mode would be NoSuchMethodError at schema-parse time, i.e. at query time. Worth aligning jackson.version to 2.14.x in a follow-up.

Verified good — evidence, not assumption

- The bump is live. avro is excluded from iceberg-core (pom.xml:249-251) and amoro-mixed-hive (:430, :488), so the direct declaration at :206-209 is the sole source. dependency:tree confirms avro:jar:1.11.4:compile. Contrast the dead guava override I found on the netty branch — this one actually takes effect.
- It fixes real exposure. 1.10.1 was affected by CVE-2024-47561 (arbitrary code execution via schema parsing, ≤1.11.3) and CVE-2023-39410. Necessary change.
- No binary incompatibility with Iceberg. I extracted the 150 classes under org/apache/iceberg/avro/* and org/apache/iceberg/data/avro/* from iceberg-core-1.7.2.jar and pulled every bytecode reference into org.apache.avro — 177 distinct members across 40 classes. All resolve against 1.11.4, and equally against 1.11.5 and 1.12.1. Zero members exist that 1.12.0 has and 1.11.x lacks on the surface Iceberg touches. This matters because iceberg-core:1.7.2 declares avro 1.12.0, so Iceberg's classes were compiled against 1.12.0 and run here against 1.11.4.
  - First pass at this gave ~35 false positives (Utf8.<init>(String) reported absent) because javap omits inherited members and my constructor parsing was wrong. Re-ran as a differential so parser flaws cancel; 16 refs filtered as artifacts, remainder clean.
  - Limits: direct bytecode references only — not reflection, service loading, or behavioral changes such as stricter schema validation.
- Trino doesn't constrain the choice. trino-hive:406 declares no avro (it uses shaded hive-apache).
- CI-equivalent build passes. Ran the exact trino-ci.yml command (-P hadoop2,format-mixed-format-trino,build-mixed-format-trino): BUILD SUCCESS, 19/19 tests, 5:29. Those tests run real queries via MixedFormatQueryRunner against keyed and Hive tables, so the Iceberg/Avro read path is genuinely exercised — which is the strongest available evidence against the Jackson concern. Javadoc plugin threw two MavenInvocationExceptions, pre-existing and non-fatal.
- Header check passes. Regex Modified by Datazip Inc\. in [0-9]{4} within head -40; added line sits at line 18.

```

PR #40, opened 2026-09-17
chore: critical vuln fixes

ashisharma-dz:

Critical Vulnerabilities

Close Alerts #43, #42, #221, #177.

## Brief change log

- bumped dependencies

# Documentation

- [ ] Documentation Link: [link to README, olake.io/docs, or olake-docs]
- [x] N/A (bug fix, refactor, or test changes only)


hash-data:
AI review 
```
P0 — Derby still ships. The CVE is not fixed.

Derby arrives transitively from org.apache.spark:spark-hive_2.13:3.5.8 (Hive metastore chain) at compile scope. Verified three ways:

amoro-ams tree:  +- org.apache.spark:spark-hive_2.13:jar:3.5.8:compile
                 |  \- org.apache.derby:derby:jar:10.14.2.0:compile

dev/deps manifests (both):        derby/10.14.2.0//derby-10.14.2.0.jar
built Trino plugin tarball:       lib/derby-10.14.2.0.jar

This also explains the commit message's "regenerated dependencies-hadoop files" producing zero manifest diff — regeneration was a no-op because nothing actually changed. The manifests still listing derby/10.14.2.0 is itself the proof the removal didn't work, and nothing in CI flags it because deps.yml is still disabled.

P1 — the change un-pinned Derby, adding a second vulnerable version

Deleting the dependencyManagement entry removed the version pin. Where everything previously resolved to 10.14.2.0, seven-plus modules now resolve derby:10.14.1.0 — older, same vulnerability:

amoro-common, amoro-format-iceberg, amoro-mixed-hive, amoro-format-mixed-spark-3-common_2.13, amoro-format-paimon, amoro-optimizer-common, amoro-optimizer-standalone, …

The shipped dist and Trino plugin still get 10.14.2.0, so nothing breaks — but a security change that adds a vulnerable version to the tree is moving backwards.

P1 — the default config still points at Derby, and its driver is now undeclared

# dist/src/main/amoro-bin/conf/config.yaml:115-118
  database:
    type: derby
    jdbc-driver-class: org.apache.derby.jdbc.EmbeddedDriver
    url: jdbc:derby:/tmp/amoro/derby;create=true

Same default in charts/amoro/values.yaml:153-155, with chart tests asserting derby behaviour (amoro-secret_test.yaml:25). AMS boots today only because spark-hive happens to drag Derby in. A documented default should not depend on an accident of someone else's transitive.

```

hash-data:
```
Verdict: do not merge. Two independent P0s. The change did not remove Derby, it downgraded it and added a new CVE, and it breaks the Helm chart with the chart's own default values.

The Derby→Postgres switch itself is the right strategic call, and the spark-hive exclusion targets exactly the dependency I traced last time. The execution is incomplete.

P0-1 — Derby is not removed. It was downgraded, and that added a CVE.

The exclusion was added to spark-hive (amoro-ams/pom.xml:157). But there is a second, independent path that was missed:

amoro-ams → amoro-mixed-hive → org.apache.hive:hive-metastore → derby

Because <derby-jdbc.version> and the dependencyManagement pin were deleted in 27c91909b, each path now resolves its own version. I regenerated both dependency manifests — this is what the build actually produces today:

dependencies-hadoop-2-spark-3.3:  committed  derby/10.14.2.0
                                  ACTUAL     derby/10.10.2.0     ← hive-metastore 2.3.8
dependencies-hadoop-3-spark-3.5:  committed  derby/10.14.2.0
                                  ACTUAL     derby/10.14.1.0     ← hive-metastore 3.1.3

So the net effect of the Derby security work:

- CVE-2022-46337 is still present in both distributions — the thing it set out to fix.
- A new CVE was introduced. Derby 10.10.2.0 (a 2014 release) is affected by CVE-2015-1832 — XXE in SqlXmlUtil, fixed in 10.12.1.1. The previously-shipped 10.14.2.0 was not affected. The hadoop-2 distribution is strictly worse than before this work started.
- The committed manifests are now wrong, claiming 10.14.2.0 while the build emits 10.10.2.0 / 10.14.1.0. deps.yml is still disabled, so nothing catches it.

The commit message "removed derby completely" is not accurate.

P0-2 — the Helm chart no longer renders with its own defaults

values.yaml now sets type: postgres but defines no username/password, while amoro-secret.yaml:30 creates a DB secret for exactly mysql or postgres:

$ helm template test charts/amoro
Error: amoro/templates/amoro-secret.yaml:32:61
  executing "amoro/templates/amoro-secret.yaml" at <b64enc>:
    invalid value; expected string

Every default helm install fails outright. Confirmed the missing credentials are the sole cause — supplying --set amoroConf.database.username=u --set amoroConf.database.password=p renders fine.

CI will not catch this. helm-charts-ci.yml runs helm unittest charts/amoro, and every case in amoro-secret_test.yaml overrides database.type (and supplies credentials for mysql/postgres). No test renders default values. The suite stays green while the chart is broken — the gap sits precisely where the default changed.

P1 — breaking change with no migration path, and a silent-data-loss mode

Existing deployments on the default Derby config get pointed at jdbc:postgresql://127.0.0.1:5432/db with username: user / password: passwd on upgrade. Two outcomes:

1. No Postgres there → AMS fails to start. Loud, recoverable.
2. A Postgres is reachable → auto-create-tables: true initialises an empty schema and AMS comes up looking healthy but completely empty — catalogs, table runtimes and optimizing state all apparently gone, while the real data sits untouched in /tmp/amoro/derby.

The second is the dangerous one. Nothing in the diff addresses migration, and there is no upgrade note. This should gate the change.

P1 — the shipped config is non-functional and carries placeholder credentials

dist/src/main/amoro-bin/conf/config.yaml:115-121 ships username: user / password: passwd against 127.0.0.1:5432/db. The out-of-box experience is now "fails to start", where it previously worked with zero configuration. If Postgres is genuinely required, that needs to be a documented prerequisite, not a placeholder that looks like it might work.
```

hash-data (approved):

